PR #54, opened 2022-09-21
move data specific info into TabularData and BindaryData message

alexis-wei:


AaronCasas (approved):
🚀

agreenb (approved):
Thanks, this is clean! 🥇

PR #58, opened 2022-09-23
Fix makefile for local builds

Otterverse:
Minor fixes for the Makefile to actually use local paths (e.g. for node modules) and install those modules. Specifically, protoc-gen-ts. The other changes are because it's now generating with the installed protoc-gen-ts and mockgen (rather than whatever system installed versions people must've previously had.)


tahiyasalam:
Just ran it locally and it worked! Thanks for the quick turnaround

tahiyasalam (approved):

PR #60, opened 2022-09-26
[DATA-475] Add mime type to captured data in API

tahiyasalam:
Add mime type metadata to binary capture, tabular capture, and user file for syncing


agreenb (approved):
Great! All of the changes to gen/js/ files is `this.hostname_ = hostname.replace(/\/+$/, '')` which I assume is from a newer version of protoc-gen-grpc-web

Before submitting, double checking that this is expected @njooma

njooma (approved):
Ya I think the generated code is fine -- *probably* nothing to worry about

PR #61, opened 2022-09-29
add labels to Geometries

bhaney:
In order to find 3D objects using the SDK, I need a label to at least switch/filter on. Given that this just adds one field to a meessage should mean its not a breaking change.


Fahmina (changes requested):
Can you add the string label to the `geometries` message type instead?

PR #69, opened 2022-10-07
App 702 - Add Creation Timestamp

nfranczak:
This pull request adds creation timestamps to: Robot, RobotPart, Organization, Location, and Fragment.


proto/viam/app/v1/app.proto
```diff
@@ -60,6 +60,7 @@ message Robot {
   string name = 2 [(tagger.v1.tags) = "bson:\"name\" json:\"name\""];
   string location = 3 [(tagger.v1.tags) = "bson:\"location\" json:\"location\""];
   google.protobuf.Timestamp last_access = 4 [(tagger.v1.tags) = "bson:\"last_access\" json:\"last_access\""];
+  google.protobuf.Timestamp created_on = 5 [(tagger.v1.tags) = "bson:\"created_on\" json:\"created_on\""];
```
AdamMagaluk, line 63: We shouldn't be adding this when the field name is the same. Of course it follows the above precedence but we should remove the json tag here. Its automatically added by the proto-gen-go.

- Same goes for all other others.

More background in: https://docs.google.com/document/d/1UYNnrx8-
15G5KjuxsO215lUKMxna6FEyecEsh55q9sA/edit

proto/viam/app/v1/app.proto
```diff
@@ -74,6 +75,7 @@ message RobotPart {
   string location_id = 12 [(tagger.v1.tags) = "bson:\"location_id\" json:\"-\""];
   google.protobuf.Struct robot_config = 5 [(tagger.v1.tags) = "bson:\"config\" json:\"robot_config\""];
   google.protobuf.Timestamp last_access = 6 [(tagger.v1.tags) = "bson:\"last_access\" json:\"last_access\""];
+  google.protobuf.Timestamp created_on = 13 [(tagger.v1.tags) = "bson:\"created_on\" json:\"created_on\""];
```
AdamMagaluk, line 78: Nit: let's move this to the end to try and keep fields in approximate field number order.  There's no specific reason it needs to be next to last_access.

AdamMagaluk (approved):
Minor nits: otherwise LGTM

proto/viam/app/v1/app.proto
nfranczak: done

PR #77, opened 2022-10-14
Update generated files: run `make dist/buf`

AdamMagaluk:
Run `make dist/buf` locally on the main branch.


stevebriskin:
Do you know where these diff originate from?
```
  this.hostname_ = hostname.replace(/\/+$/, '');
  this.hostname_ = hostname;
```

stevebriskin (approved):

AdamMagaluk:
I think a difference in tooling. But i cannot tell if it's a difference only on my machine or because we bumped the version of our tools. If you run `make dist/buf ` locally on the main branch do you get the same changes?

PR #80, opened 2022-10-24
[RSDK-830] add DependsOn and models to serviceConfig

Kschappacher:
Needed change for the adding dependencies to Service change: https://github.com/viamrobotics/rdk/pull/1518


Kschappacher:
I ran dist/buf locallyI think the CI is a bit buggy rn

proto/viam/app/v1/robot.proto
```diff
@@ -81,6 +81,7 @@ message ServiceConfig {
   string namespace = 2;
   string type = 3;
   google.protobuf.Struct attributes = 4;
+  repeated string depends_on = 5;
```
cheukt, line 84: we need to add models here too - maybe add it in this PR and then make a separate ticket for the associated changes in rdk/app

cheukt (approved):

PR #84, opened 2022-10-26
[RSDK-556] remove width/height from camera GetImage response

gvaradarajan:
https://github.com/viamrobotics/rdk/pull/1447 to be released immediately after


njooma (approved):

.github/workflows/pullrequest.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       options: --platform linux/amd64
     steps:
     - uses: actions/checkout@v2
-    - uses: bufbuild/buf-setup-action@v1
+    - uses: bufbuild/buf-setup-action@v1.8.0
```
njooma, line 25: Do we need to pin this? If yes, let's do 1.9.0

gvaradarajan: no I just forgot to undo that

PR #87, opened 2022-10-31
[DATA-576] Switch to multiple location IDs in filter

agreenb:
This is a change that only breaks the UI and is handled/fixed in the corresponding app change: https://github.com/viamrobotics/app/pull/825


AaronCasas (approved):
LGTM

tahiyasalam (approved):

PR #88, opened 2022-11-01
[RSDK-812-817] Add extra param to services

cheukt:
adds extra struct to all services. Will not be merged until the corresponding PRs on the RDK/Python SDK is ready


njooma (approved):
LGTM. Since this is mostly additive, we can probably merge this in without waiting on SDK updates

PR #93, opened 2022-11-03
[Data-700] extra params for audio input

bhaney:
adds extra field to all audio input methods


edaniels (approved):

cheukt (approved):

bhaney:
won't do this right now per discussion with Eric and Cheuk

PR #98, opened 2022-11-07
[Data-604] camera extra params

bhaney:
Adds the "extra" to camera proto


bhaney:
won't do this right now per discussion with Eric and Cheuk

PR #99, opened 2022-11-07
Update setup to install go 1.19

maximpertsov:
Mirrors change in RDK https://github.com/viamrobotics/rdk/commit/c61aeaaa66ec76bc30c5665ea8690e8a23b01727

### DIFF
--- a/etc/setup.sh
+++ b/etc/setup.sh
@@ -154,7 +154,7 @@ do_brew(){
 
 	# pinned
 	brew "gcc@11"
-	brew "go@1.18"
+	brew "go@1.19"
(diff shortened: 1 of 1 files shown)


Otterverse (changes requested):
There were a few other changes made in setup.sh along with go 1.19 (bugs in the syntax of the .viamdevrc and around the ordering of installation with brew.) It'd be good to include all of them so a setup here doesn't break RDK. Maybe just copy the current setup.sh from RDK itself?

Otterverse (approved):
LGTM

PR #104, opened 2022-11-10
Update the github token

njooma:
This is a follow up to https://github.com/viamrobotics/api/pull/101, which should fix this https://github.com/viamrobotics/api/actions/runs/3438079708

The issue is that the main branch is protected (as it should be). Can an admin @edaniels @Otterverse etc. update the permissions on this so that the `viambot` user can push directly to main? 

This PR updates the GitHub tokens to use the `viambot` token, similar to how I do it for the [python-sdk](https://github.com/viamrobotics/viam-python-sdk/blob/main/.github/workflows/release.yml#L34), which similarly needs to push directly to main

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -24,6 +24,8 @@ jobs:
       options: --platform linux/amd64
     steps:
       - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.REPO_READ_TOKEN }}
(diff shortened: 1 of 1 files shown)


Otterverse:
Curious that the REPO_READ_TOKEN allows write access, but I'll document that in confluence. I've updated the branch protection to allow viambot to bypass the PR requirement.

Otterverse (approved):
LGTM

edaniels (approved):

PR #107, opened 2022-11-11
Allow action to force push

njooma:


edaniels:
@njooma I think you may want to update rdk/app to mention that we are now pinned to API versions and must explicitly ask for them (particularly for web assets)

PR #109, opened 2022-11-16
[DATA-671] Refactor data sync to be unary instead of a bidi stream.

AaronCasas:


edaniels:
Please update PR title

AaronCasas:
@tessavitabile wanted to ask your opinion on how to handle this breaking change!

For this change, I see two options:
- Make this api change and this [app](https://github.com/viamrobotics/app/pull/921) change, deleting the old rpcs. This is sort of maximally breaking: any existing, not updated robot trying to sync things will get start getting unimplemented errors from the backend, because the old rpcs they're trying to use have been removed. This will occur until they update rdk on their robot.
- Don't delete either of these for some cutover period. The old rpcs and these new ones will both exist for this cutover period, so existing syncing robots will not be impacted. Downside here is that we'll have a whole lot of deprecated code around for this cutover period, and I'll need to spend some time breaking up that PR.

Is there a strong preference at this point for the latter vs the former? I know long term we should avoid this sort of thing, but afaik Klare is the only active user of data sync, so we can coordinate with her when making this change. I saw you posted recently about breaking change practices at Mongo, so wanted to run it by you

proto/viam/app/datasync/v1/data_sync.proto
```diff
@@ -10,9 +10,26 @@ import "tagger/v1/tagger.proto";
...
 
+message DataCaptureUploadRequest {
+  UploadMetadata metadata = 1;
+  repeated SensorData sensor_contents = 2;
+}
+
+message DataCaptureUploadResponse {}
+
+message FileUploadRequest {
+  oneof upload_packet {
```
tessavitabile, line 25: @edaniels, are we supposed to be avoiding `oneof` in requests as well as responses?

edaniels:
So generally we would need to support both APIs. I think the question we should ask ourselves and find the answer to is: who and how many people are currently using the old API? If it's non-zero, is it easy enough to coordinate the breaking change? If not, then we may want to support both. Breaking changes being approved is less a me or James thing, but more a product decision to make (we just have admin privs to merge).

In our current stage of development, if we knew that a feature had low use and it makes our lives easier to do a clean break and notify the person while hand holding them, I'd opt for it.

proto/viam/app/datasync/v1/data_sync.proto
edaniels: I think if oneof can be avoided easily, it should. In this case it seems okay to have a stream use a oneof since it represents stages. It needs comments though as to how to use it properly.

agreenb: +1, and we'll reuse these comments in the PackageService's CreatePackageRequest oneof.

agreenb (approved):
Since the corresponding RDK and app PRs are approved or in the last review, approving here. 

Before submitting this Monday morning, request to send an update in #eng-breaking-changes. Once it's all merged, would test on a robot with the latest version of RDK for syncing tabular data, binary data, file data, and viewing all of them as expected in our app UI. And we'll reach out to Klare and Joe (currently the folks using it) directly through the process, which'll also be helpful for feedback.

PR #115, opened 2022-11-22
missed built go files

sbal13:
Some of the go files weren't built in the last dist/buf


alexis-wei (approved):
I hope these are right

PR #124, opened 2022-12-02
Workflows to generate and commit protos

njooma:
GitHub doesn't allow workflows to access secrets on PRs from forks (https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). So previously when compiling protos, the workflow could not commit the code back to the head branch because it did not have the necessary secret to do so. 

The new workflow is thus:

1. On PR approval, run an INSECURE workflow that saves the PR information to be used in the next step. This was borrowed from the above link
2. On successful run of the PR Approval workflow, a SECURE workflow kicks off that can compile protos and commit them back in. However, because this workflow is no longer associated with the PR, having it as a required check for merging will never pass. So this workflow then labels the PR with a `protos-compiled` label
3. A third workflow that is associated with the original PR sits waiting for label changes. When a label is added/removed, this workflow will run and can be used as a required check: if the necessary labels exist (currently only `photos-compiled`), then it will pass. Otherwise, it will fail.

The workflows in order are:
1. on_pr_approved.yml
2. compile_protos.yml
3. check_merge_reqs.yml

### DIFF
--- a/.github/workflows/check_merge_reqs.yml
+++ b/.github/workflows/check_merge_reqs.yml
@@ -1,0 +1,27 @@
+name: Check Requirements for Merge
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    types:
+      - labeled
+      - unlabeled
+
+jobs:
+  check_requirements:
+    name: Check Merge Requirements
+    runs-on: [self-hosted, x64]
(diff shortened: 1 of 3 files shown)


edaniels (approved):
Impressive

PR #126, opened 2022-12-07
Support JWKS auth config

AdamMagaluk:
Add a `CredentialsType_CREDENTIALS_TYPE_WEB_OAUTH` type as a valid AuthHandler. This RPC handler will soon be supported in the RDK that will give the RDK the same ability to accept access tokens provided by our auth system for the web app. It accepts a generic [JSON Web Tokens File (JWKS)](https://www.rfc-editor.org/rfc/rfc7517) as well as a list of valid audiences.

In addition `AuthHandlerConfig` is extended with a `web_oauth_config` config field. This is used when the config type of the auth handler is CREDENTIALS_TYPE_WEB_OAUTH. In that case the generic struct config is ignored.


proto/viam/app/v1/robot.proto
```diff
@@ -108,9 +108,25 @@ message AuthConfig {
...
+  repeated string allowed_audiences = 1;
+  JWKSFile jwks = 2;
+}
+
 message AuthHandlerConfig {
   CredentialsType type = 1;
   google.protobuf.Struct config = 5;
+
+  // Structured config for CREDENTIALS_TYPE_WEB_OAUTH type. Note: When this is set the config field will be empty.
+  optional AuthHandlerWebOauthConfig web_oauth_config = 6;
```
edaniels, line 129: in v2 we will make a better config

proto/viam/app/v1/robot.proto
```diff
@@ -108,9 +108,25 @@ message AuthConfig {
...
 
+message JWKSFile {
+  // JSON Web Keys (JWKS) file as arbitary json.
+  // See https://www.rfc-editor.org/rfc/rfc7517
+  google.protobuf.Struct json = 1;
+}
+
+// Structured config for the credential type CREDENTIALS_TYPE_WEB_OAUTH
+message AuthHandlerWebOauthConfig {
+  // The allowed aud claimms by the web oauth handler.
```
edaniels, line 119: claims

edaniels (approved):

edaniels (approved):

PR #130, opened 2022-12-13
[DATA-835] Add PackageService with Create/Delete/Get/ListPackage endpoints to API

tahiyasalam:


proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  // DeletePackage receives an authenticated request with the package name and array of versions and removes the associated packages
+  rpc DeletePackage(viam.app.package.v1.DeletePackageRequest) returns (viam.app.package.v1.DeletePackageResponse) {
+    option (google.api.http) = {
+      delete: "/package/v1/delete_package"
+    };
+  }
+
+  // GetPackages receives an authenticated request with the package name and version and, if present, generates returns the URL and metadata
```
tahiyasalam, line 28: @agreenb wondering if comments, such as " // Primarily used by robot to download package files." from the scope doc should be included here

agreenb: Good question but nope, those were mostly used to describe the intent of each part of the scope.

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  // DeletePackage receives an authenticated request with the package name and array of versions and removes the associated packages
+  rpc DeletePackage(DeletePackageRequest) returns (DeletePackageResponse) {
+    option (google.api.http) = {
+      delete: "/viam/api/v1/service/datamanagement/{organization_id}/delete"
+    };
+  }
+
+  // GetPackages receives an authenticated request with the package name and version and, if present, generates returns the URL and metadata
```
agreenb, line 25: Since these all require authenticated requests, think can remove. Suggestion:

`// GetPackage returns the URL and metadata for a requested package version.`

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+
+package viam.app.package.v1;
+
+import "google/api/annotations.proto";
+import "google/protobuf/struct.proto";
+
+option go_package = "go.viam.com/api/app/package/v1";
+
+service PackageService {
+  // CreatePackage uploads a package's contents and information to the cloud
```
agreenb, line 11: I think this can just be `// CreatePackage uploads a package to the cloud`

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  rpc CreatePackage(stream CreatePackageRequest) returns (CreatePackageResponse) {
+    option (google.api.http) = {
+      post: "/viam/api/v1/service/datamanagement/create"
+    };
+  }
+
+  // DeletePackage receives an authenticated request with the package name and array of versions and removes the associated packages
```
agreenb, line 18: `// DeletePackage removes the given package versions.`

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  optional string name = 2;
+  optional string version = 3;
+}
+
+message GetPackageResponse {
+  repeated Package packages = 1;
+}
+
+message ListPackagesRequest {
+  string organization_id = 1; // If only organization_id is specified, get all packages for an organization. Used in app.
```
agreenb, line 95: Can remove this comment and the one below.

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  // GetPackages receives an authenticated request with the package name and version and, if present, generates returns the URL and metadata
+  rpc GetPackage(GetPackageRequest) returns (GetPackageResponse) {
+    option (google.api.http) = {
+      get: "/viam/api/v1/service/datamanagement/{organization_id}/package"
+    };
+  }
+
+  // ListPackages receives an authenticated request with the organization ID and other optional fields and, if present, generates returns the URL and metadata of all associated packages
```
agreenb, line 32: Something like: `// ListPackages get the URLs and metadata for requested packages. Include package name, version, and/or type to filter beyond the required organization_id.`

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  rpc ListPackages(ListPackagesRequest) returns (ListPackagesResponse) {
+    option (google.api.http) = {
+      get: "/viam/api/v1/service/datamanagement/{organization_id}/list"
+    };
+  }
+}
+
+enum PackageType {
+  PACKAGE_TYPE_UNSPECIFIED = 0;
+  PACKAGE_TYPE_FILE = 1;
```
agreenb, line 42: Updated previously in scope to PACKAGE_TYPE_ARCHIVE based on feedback to represent a tar archive of files

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  string version = 3;
+  PackageType type = 4;
+  repeated FileInfo files = 5;
+  google.protobuf.Struct metadata = 6;
+}
+
+message CreatePackageRequest {
+  oneof package {
+    PackageInfo info = 1;
+    bytes contents = 2; // .tar.gz file with a limit of 50GB
```
agreenb, line 63: Can remove the limit from here since that'll be a variable that could be changed on the BE. I think ok to leave .tar.gz file comment in here since might be helpful just by looking at the API docs.

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  rpc CreatePackage(stream CreatePackageRequest) returns (CreatePackageResponse) {
+    option (google.api.http) = {
+      post: "/viam/api/v1/service/datamanagement/create"
+    };
+  }
+
+  // DeletePackage receives an authenticated request with the package name and array of versions and removes the associated packages
+  rpc DeletePackage(DeletePackageRequest) returns (DeletePackageResponse) {
+    option (google.api.http) = {
+      delete: "/viam/api/v1/service/datamanagement/{organization_id}/delete"
```
agreenb, line 21: Are these autogenerated? if not, think these {organization_id}'s can be removed here and below since organization is required but the methods involve a number of other required fields

agreenb (changes requested):
Thanks for combining all of this into one and reflecting today's final changes to the doc!

agreenb:
Mostly just nit suggestions for shorter gRPC method descriptions. Nice work!

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done! thanks for letting me know

proto/viam/app/package/v1/package.proto
tahiyasalam: done!

proto/viam/app/package/v1/package.proto
tahiyasalam: done~

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+import "google/api/annotations.proto";
+import "google/protobuf/struct.proto";
+
+option go_package = "go.viam.com/api/app/package/v1";
+
+service PackageService {
+  // CreatePackage uploads a package to the cloud
+  rpc CreatePackage(stream CreatePackageRequest) returns (CreatePackageResponse) {
+    option (google.api.http) = {
+      post: "/viam/app/package/v1/create"
```
agreenb, line 14: @sbal13 Would be useful to get your feedback on these names for consistency with typical REST endpoints. 

@tahiyasalam originally suggested :organization_id here but there will always be an org, package name, and package version, which seems like too much to include :)

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  rpc GetPackage(GetPackageRequest) returns (GetPackageResponse) {
+    option (google.api.http) = {
+      get: "/viam/app/package/v1/package"
+    };
+  }
+
+  // ListPackages get the URLs and metadata for requested packages. Include package name, version, and/or type to filter beyond the required organization_id.
+  rpc ListPackages(ListPackagesRequest) returns (ListPackagesResponse) {
+    option (google.api.http) = {
+      get: "/viam/app/package/v1/list"
```
agreenb, line 35: If the above remains v1/package, would switch this to v1/packages.

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+}
+
+message GetPackageRequest {
+  string organization_id = 1;
+  optional string name = 2;
+  optional string version = 3;
+}
+
+message GetPackageResponse {
+  repeated Package packages = 1;
```
agreenb, line 91: `Package package = 1;`

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+message DeletePackageResponse {
+  int64 deleted_count = 1; // Number of versions deleted
+}
+
+message Package {
+  PackageInfo info = 1;
+  string url = 2;
+}
+
+message GetPackageRequest {
```
agreenb, line 84: These will all be required and return a single package, whereas ListPackages will allow you to get multiple.

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  repeated string versions = 3;
+}
+
+message DeletePackageResponse {
+  int64 deleted_count = 1; // Number of versions deleted
+}
+
+message Package {
+  PackageInfo info = 1;
+  string url = 2;
```
agreenb, line 81: [supernit] Not in the doc, but would switch over to URI for consistency with our other APIs (a superset of URL)

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+message CreatePackageResponse {}
+
+message DeletePackageRequest {
+  string organization_id = 1;
+  string name = 2;
+  repeated string versions = 3;
+}
+
+message DeletePackageResponse {
+  int64 deleted_count = 1; // Number of versions deleted
```
agreenb, line 76: [supernit] can put the comment above the field

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,103 @@
...
+  string version = 3;
+  PackageType type = 4;
+  repeated FileInfo files = 5;
+  google.protobuf.Struct metadata = 6;
+}
+
+message CreatePackageRequest {
+  oneof package {
+    PackageInfo info = 1;
+    bytes contents = 2; // .tar.gz file
```
agreenb, line 63: [supernit] Can put the comment above the field, since we'll update this comment in a follow-up PR to match the wording in [data_sync.proto](https://github.com/viamrobotics/api/pull/109#discussion_r1047683742)

agreenb (changes requested):
Just a few last nit things, then looks great. Added @sbal13 to look at the REST apis

proto/viam/app/package/v1/package.proto
tahiyasalam: done

proto/viam/app/package/v1/package.proto
tahiyasalam: ah, right done

proto/viam/app/package/v1/package.proto
tahiyasalam: done

proto/viam/app/package/v1/package.proto
tahiyasalam: done

proto/viam/app/package/v1/package.proto
tahiyasalam: done

proto/viam/app/package/v1/package.proto
sbal13: Eric Adam and I talked about this and decided that the viam and app bits are unnecessary, though you'll see this format in other protos that were made before this. Basically we decided it should be `/<service>/<verison>/<action` so in this case just the `/package/v1/create` bit

sbal13: we're not necessarily going for REST conventions here, more just trying to make them expressive of the services/functions that these endpoints are proxies for

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,105 @@
...
+  rpc DeletePackage(DeletePackageRequest) returns (DeletePackageResponse) {
+    option (google.api.http) = {
+      delete: "/viam/app/package/v1/delete"
+    };
+  }
+
+  // // GetPackage returns the URL and metadata for a requested package version
+  rpc GetPackage(GetPackageRequest) returns (GetPackageResponse) {
+    option (google.api.http) = {
+      get: "/viam/app/package/v1/package"
```
sbal13, line 28: nit: maybe `/package/v1/get`?

proto/viam/app/package/v1/package.proto
```diff
@@ -0,0 +1,105 @@
...
+    };
+  }
+
+  // ListPackages get the URLs and metadata for requested packages. Include package name, version, and/or type to filter beyond the required organization_id.
+  rpc ListPackages(ListPackagesRequest) returns (ListPackagesResponse) {
+    option (google.api.http) = {
+      get: "/viam/app/package/v1/packages"
```
sbal13, line 35: nit: maybe `/package/v1/list`? open to whatever for these, just trying to stick to a convention where the last bit is a verb like 'create' or 'delete'

agreenb (commented):
LGTM after Steven's suggestions. I think it'll require another re-approval after any commit so holding off until after. ✅✅

proto/viam/app/package/v1/package.proto
tahiyasalam: totally, makes sense!

proto/viam/app/package/v1/package.proto
tahiyasalam: done :) thanks!

proto/viam/app/package/v1/package.proto
tahiyasalam: cool!

agreenb: Thanks Steven - @tahiyasalam I think this all looks great except seems like can get rid of the /viam/app prefix!

tahiyasalam: my bad! should be fixed now

agreenb (approved):
🚀🚀🚀
Thanks for adding!

PR #131, opened 2022-12-13
add sharelocation + unsharelocation apis

michaellee1019:


proto/viam/app/v1/app.proto
```diff
@@ -139,6 +145,20 @@ message ListLocationsRequest {
   string organization_id = 1;
 }
 
+message ShareLocationRequest {
+  string location_id = 1;
+  string organization_id = 2;
```
AdamMagaluk, line 150: Nit: Please add some docs on these fields. Eg. it would be good to indicate organization_id is org that it will be shared with.

AdamMagaluk (approved):
LGTM. Just nit about docs the more the better.

AdamMagaluk (approved):

PR #123, opened 2022-12-02
Update app protos

DTCurrie:
Opening this as a draft for team review before opening this for full review. This is a work in progress and will be updated.

PR #135, opened 2022-12-20
Get/Update/Delete Location + List Robots APIs

michaellee1019:
Adds new protos for Get/Update/Delete Location as well as List Robots.

After refactoring of our app, FindRobots will be removed in favor of ListRobots that follows our desired naming conventions.

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -145,45 +145,59 @@ message ListOrganizationsResponse {
 }
 
 message Location {
+  // Location ID.
   string id = 1;
+  // Location name.
   string name = 2;
+  // Location creation timestamp.
   google.protobuf.Timestamp created_on = 3;
 }
 
 message CreateLocationRequest {
+  // Organization ID to create the location under.
   string organization_id = 1;
+  // Name of the location.
   string name = 2;
 }
 
 message CreateLocationResponse {
+  // Location object is returned.
   Location location = 1;
 }
 
 message GetLocationRequest {
+  // Location ID of location to get.
   string location_id = 1;
 }
 
 message GetLocationResponse {
+  // Location object is returned.
   Location location = 1;
 }
 
 message UpdateLocationRequest {
+  // Location ID of location to update.
   string location_id = 1;
-  string name = 2;
+  // The new to be updated on location.
+  optional string name = 2;
+  // The new parent location to move the location under.
   optional string parent_location_id = 3;
 }
 
 message UpdateLocationResponse {
+  // Location object is returned.
   Location location = 1;
 }
 
 message DeleteLocationRequest {
+  // Location ID of location to delete.
   string location_id = 1;
 }
 
 message DeleteLocationResponse {}
 
 message ListLocationsRequest {
+  // Organization ID under which to list all locations.
   string organization_id = 1;
 }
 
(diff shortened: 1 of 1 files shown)


proto/viam/app/v1/app.proto
```diff
@@ -127,21 +145,59 @@ message ListOrganizationsResponse {
...
+message GetLocationResponse {
+  // Location object is returned.
+  Location location = 1;
+}
+
+message UpdateLocationRequest {
+  // Location ID of location to update.
+  string location_id = 1;
+  // The new to be updated on location.
+  optional string name = 2;
```
michaellee1019, line 182: I changed name to be `optional` which is different from what we reviewed in https://github.com/viamrobotics/api/pull/123/files#diff-e842ca0f50219c67d40be90a2a1efd40ee1b6df19a254357ab177d631b3ea4b4R313

I figure that during an update operation, most commonly either `name` or `parent_location_id` will be updated, but also updating both at the same time will be supported.

proto/viam/app/v1/app.proto
```diff
@@ -144,20 +144,72 @@ message ListOrganizationsResponse {
...
+
+message LocationOrganization {
+  // Organization ID the location is shared with.
+  string organization_id = 1;
+  // Whether the organization is the primary owner or not.
+  bool primary = 2;
+}
+
+message LocationAuth {
+  // Deprecated: use secrets field.
```
AdamMagaluk, line 158: I think we should just drop this.

proto/viam/app/v1/app.proto
```diff
@@ -144,20 +144,72 @@ message ListOrganizationsResponse {
...
+  repeated SharedSecret secrets = 3;
+}
+
 message Location {
   // Location ID.
   string id = 1;
   // Location name.
   string name = 2;
+  // Location ID of the parent location.
+  string parent_location_id = 3;
```
AdamMagaluk, line 174: We shouldn't reuse field numbers. Go to 4.

proto/viam/app/v1/app.proto
michaellee1019: Do you know of a way to delete the field and get the buf-breaking checks to pass? I tried to add `reserved`. Any other way to ignore/bypass the check since we want it deleted?

michaellee1019: As we discussed, adding the field back for now :)

proto/viam/app/v1/app.proto
AdamMagaluk: 👍

proto/viam/app/v1/app.proto
```diff
@@ -114,7 +132,10 @@ message RobotPartHistoryEntry {
   RobotPart old = 4 [(tagger.v1.tags) = "bson:\"old\" json:\"old\""];
 }
 
-message ListOrganizationsRequest {}
+message ListOrganizationsRequest {
+  // Returns only organizations that are shared with a specific Location ID.
+  optional string shared_with_location_id = 3;
```
AdamMagaluk, line 137: This feels funky personally. Did you discuss this elsewhere?

michaellee1019: Yeah it was a tradeoff. I will send you the slack thread about it.

michaellee1019: ok. Updated to remove this field as we will handle in our frontend.

AdamMagaluk (approved):

michaellee1019:
Thanks for the review Adam!

PR #137, opened 2022-12-21
[RSDK-1597] Update package proto

agreenb:
Update to a `packages` package and directory.


agreenb:
@stevebriskin Attempted to add the linter into our toolkit, but wasn't quickly able to get it working. I assume the java issue is with the namespace rather than the RPC service, and there's no requirement for the directory/package to match the service name. 

Think PackageService is still preferable to PackagesService.

stevebriskin (approved):

agreenb:
+ @sbal13 for the gateway annotations

sbal13 (approved):
🔥

tahiyasalam (approved):

tahiyasalam (commented):
Quick check: under app and gen/js/app, we have both package/ and packages/

Maybe remove and run `make/dist-buf` again?

agreenb:
@stevebriskin This is a breaking change since removes the old "package.proto" file, so good to force merge!

Double checking again that the issue was the namespace and not `message Package`, which is currently still present.

stevebriskin:
Yes, the issue was with the package namespace, not the service name -- `package viam.app.package.v1;`

agreenb:
@tahiyasalam Great catch, wiped gen/js clean locally to get it working.

@stevebriskin now ready :)

PR #138, opened 2022-12-21
[RSDK-1596] Add Stop Signal/Timeout fields to ProcessConfig

Otterverse:
This supports the two new optional fields in goutil's pexec.ProcessConfig


proto/viam/app/v1/robot.proto
```diff
@@ -88,6 +88,8 @@ message ProcessConfig {
   string cwd = 4;
   bool one_shot = 5;
   bool log = 6;
+  int32 stop_signal = 7;
```
edaniels, line 91: opt: makeit an enum

proto/viam/app/v1/robot.proto
```diff
@@ -88,6 +88,8 @@ message ProcessConfig {
...
+  int64 stop_timeout = 8;
```
edaniels, line 92: seems like you should use proto duration for this

edaniels (changes requested):

proto/viam/app/v1/robot.proto
Otterverse: I'd rather not enum it at the proto level. On a few weird architectures (e.g. if we ever supported some MIPS or SPARC board for some reason), some of the signals are different, so an int would still let something pass the "true" signal without needing a proto change, and the enum could filter at the application level.

proto/viam/app/v1/robot.proto
Otterverse: Good call... I didn't know that was a thing (I knew about timestamp, but didn't realize specifically duration was available.)

proto/viam/app/v1/robot.proto
edaniels: good point.

edaniels (approved):

PR #139, opened 2022-12-22
[RSDK-1598] adds created on timestamp

sbal13:
The ticket mentions setting the timestamp as `5` but this causes an issue in the CI because it's detected as a breaking change (since the fields at 5 and 6 had to be shifted down)

image


agreenb (approved):
💯

alexis-wei (approved):

PR #142, opened 2022-12-28
[DATA-837] Changes for implement CreatePackage in app Package Service

tahiyasalam:
`PackageInfo` will be sent in the upload request by the user
`UploadedPackage` will be saved in the DB 
`Package` will be returned to the user in the GetPackage(s) response(s) 

Associated `app` changes are [here](https://github.com/viamrobotics/app/pull/1009)


proto/viam/app/packages/v1/packages.proto
```diff
@@ -56,7 +57,6 @@ message PackageInfo {
   PackageType type = 4;
   repeated FileInfo files = 5;
   google.protobuf.Struct metadata = 6;
-  google.protobuf.Timestamp created_on = 7;
```
agreenb, line 59: I would remove this in a separate PR since it's a breaking change.

agreenb (changes requested):
Looks great, and agree that it's better to embed the fields directly in the UploadedPackage.

proto/viam/app/packages/v1/packages.proto
```diff
@@ -83,6 +83,18 @@ message DeletePackageResponse {
 message Package {
   PackageInfo info = 1;
   string uri = 2;
+  google.protobuf.Timestamp created_on = 3;
+}
+
+message UploadedPackage {
```
agreenb, line 89: Can you add a comment that this is what's stored in the DB?

proto/viam/app/packages/v1/packages.proto
tahiyasalam: Yup, added a ticket as a reminder

proto/viam/app/packages/v1/packages.proto
AaronCasas: If the difference between Package and UploadedPackage is essentially that UploadedPackage is our internal view, I wonder if it should have a name indicating that. Might be worth asking Eric if there's a generally accepted pattern to follow here? Non blocking, but worth checking imo

tahiyasalam: done!

agreenb: +1 I think InternalPackage would be a good rename, and include that wording of "our internal view" in the comment.

We'll include Eric in the force-merge to remove created_on so can additionally get input there — want this merge to be unblocked for now.

tahiyasalam: updated

agreenb (approved):
Looks great, thanks for making the updates!

PR #143, opened 2022-12-29
[DATA-975] Add proto definition for MLTrainingService

AaronCasas:
Also adds messages needed for TrainingJobManager cron.

Ticket: https://viam.atlassian.net/browse/DATA-975
Spec: https://docs.google.com/document/d/1cZzoaMdsm226LtDsxawxop4Oh384OC_82BXFYjMK8Yw/edit#


proto/viam/app/mltraining/ml_training.proto
```diff
@@ -0,0 +1,68 @@
...
+package viam.app.mltraining.v1;
+
+import "app/data/v1/data.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "go.viam.com/api/app/mltrainining/v1";
+
+service MLTrainingService {
+  rpc SubmitTrainingJob(SubmitTrainingJobRequest) returns (SubmitTrainingJobResponse);
```
AaronCasas, line 12: As we add the UI, I think there are additional rpcs we're going to want here, such as GetTrainingJobs(orgid) GetRunningJobs(orgid) etc. I think we should wait until that's being worked on to add them so we actually know what we want/need, but calling that out

proto/viam/app/mltraining/ml_training.proto
```diff
@@ -0,0 +1,68 @@
...
+message SubmitTrainingJobResponse {
+  string id = 1;
+}
+
+message GetTrainingJobRequest {
+  string id = 1;
+}
+
+message GetTrainingJobResponse {
+  string id = 1;
```
tahiyasalam, line 40: nit/q: Do we need id in both the request and response?

proto/viam/app/mltraining/ml_training.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
+
+package viam.app.mltraining.v1;
+
+import "app/data/v1/data.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "go.viam.com/api/app/mltrainining/v1";
```
tahiyasalam, line 9: nit: training spelled wrong here

tahiyasalam: also: do we want to have this file in a folder called `v1` similar to datasync, data, etc..

AaronCasas: oops yes 100%, I accidentally made an empty v1 folder and left this outside of it, moved

proto/viam/app/mltraining/ml_training.proto
AaronCasas: So interestingly enough this is not a settled question and there's a lot of room for debate. Relevant [SO](https://stackoverflow.com/questions/11299354/should-a-restful-get-response-return-a-resources-id) (REST vs rpc, but this is a GET method for a resource so seems appropriate)

I personally lean towards including it as a general rule, because then the response can be referenced/used other than in the immediate context of when making the request. So for example if you have this response and want to refresh that, you can do that, even if you didn't save the id when you made the initial request.

We obviously don't do this in most places, though. If you think that inconsistency makes it confusing happy to remove too, I don't feel strongly

tahiyasalam: Fine to keep, makes sense to me in the context of refreshing

tahiyasalam (approved):
Neat, looks good!

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -21,7 +21,7 @@ enum ModelType {
 
 message SubmitTrainingJobRequest {
   viam.app.data.v1.Filter filter = 1;
-  string organization_id = 2;
+  string org_id = 2;
```
tahiyasalam, line 24: Should actually be organization_id as we're moving to organization_id instead of org_id for consistency across packages

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -0,0 +1,68 @@
...
+option go_package = "go.viam.com/api/app/mltraining/v1";
+
+service MLTrainingService {
+  rpc SubmitTrainingJob(SubmitTrainingJobRequest) returns (SubmitTrainingJobResponse);
+  rpc GetTrainingJob(GetTrainingJobRequest) returns (GetTrainingJobResponse);
+}
+
+enum ModelType {
+  MODEL_TYPE_UNSPECIFIED = 0;
+  MODEL_TYPE_MULTI_CLASS_CLASSIFICATION = 1;
```
agreenb, line 18: After talking this through with design, I realize this should be SINGLE_LABEL_CLASSIFICATION and MULTI_LABEL_CLASSIFICATION. Multi-class inherently means > 1 classes, whereas I was assuming it was 1+ classes. See the breakdown [here](https://cloud.google.com/vertex-ai/docs/image-data/classification/prepare-data)

For either single- or multi-label, there could be multiple possible labels as inputs (dog, cat) but the above distinctions determines whether we'd only predict one (dog or cat) or multiple (dog and cat) per image.
For the binary classification case, the user would only pick one input label (dog) and single label.

Will update accordingly in the scope if have agreement here, cc @tahiyasalam

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -0,0 +1,68 @@
...
+  MODEL_TYPE_MULTI_LABEL_CLASSIFICATION = 2;
+}
+
+message SubmitTrainingJobRequest {
+  viam.app.data.v1.Filter filter = 1;
+  string org_id = 2;
+  string model_name = 3;
+  string model_version = 4;
+  ModelType model_type = 5;
+  repeated string training_labels = 6;
```
agreenb, line 28: Since this can be expanded later to non-string labels, I think should remain consistent with our API and switch to `repeated string tags` 

The other option is to make this a nested oneof Label type, but decided against that in our existing API for 1) simplicity and 2) since tags will likely be the most common label type.

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -0,0 +1,68 @@
...
+
+enum TrainingStatus {
+  TRAINING_STATUS_UNSPECIFIED = 0;
+  TRAINING_STATUS_PENDING = 1;
+  TRAINING_STATUS_IN_PROGRESS = 2;
+  TRAINING_STATUS_COMPLETED = 3;
+  TRAINING_STATUS_FAILED = 4;
+}
+
+message TrainingJobMetadata {
```
agreenb, line 52: Would ideally store the tags that we trained on, enabling the user to easily retrain on the same tags if retraining/updating an existing model.

agreenb (commented):
Clean add! A few comments.

proto/viam/app/mltraining/v1/ml_training.proto
AaronCasas: Sounds good, changed

proto/viam/app/mltraining/v1/ml_training.proto
AaronCasas: Updated to tags. 

Is the idea that in the future we'd deprecate this tags field in favor of some more general label field? Or would it be in addition to it?

proto/viam/app/mltraining/v1/ml_training.proto
AaronCasas: Those tags are in the `SubmitTrainingJobRequest request` field, so will still be saved/queryable

proto/viam/app/mltraining/v1/ml_training.proto
agreenb: Thinking through. In the Data API it makes sense to keep separate as there could be different types of labels for different purposes on a single datum, and the user will usually want to grab all tags _or_ bounding boxes _or_ segmentation masks for each distinct model type, whereas here we'd likely pass one type in.

We've also been careful about using oneofs which are a bit more clunky to set/get/understand — vs just listing all the fields — but otherwise I'd see this as equally valid:

```
message Labels {
   oneof {
        repeated string tags = 1; 
   }
}
```

AaronCasas: I'm gonna ask Eric about the oneof thing, I've never really gotten why we avoid them as a general rule.

In that case though it sounds like we'd start with just tags, but presumably deprecate that field in the medium term future in favor of some `Labels` once we add the ability to train models on different label types (+ add the ability to add those labels)?

proto/viam/app/mltraining/v1/ml_training.proto
AaronCasas: @tahiyasalam just double checking - do you agree?

tahiyasalam: Oops, sorry missed the sentence about agreement and thought you were just cc'ing me.

But yes, I agree!

proto/viam/app/mltraining/v1/ml_training.proto
agreenb: SGTM!

agreenb (approved):
🚀

PR #144, opened 2023-01-03
[DATA-897] Remove unused `last` from BinaryDataByIDsResponse

agreenb:
Cleanup PR for a breaking API change to remove a field that was never used.

`last` is used for [pagination](https://github.com/viamrobotics/api/blob/429230e61cfdb32e1ce8f5aee539f8422c066d24/proto/viam/app/data/v1/data.proto#L67-L68) when requesting data by filter in Binary/TabularDataByFilterRequest.

We also support requesting binary data by IDs, but this doesn't include any pagination in the request and thus `last` is unused.

Removed in app [here](https://github.com/viamrobotics/app/pull/1026/files) (which was simply populating the field with an empty string) and was never used in RDK (which [only uses the `data` field](https://github.com/viamrobotics/rdk/blob/0dc8464545a1d7f7881021aed44869df7561c199/cli/data.go#L170-L181) of BinaryDataByIDsResponse)


tahiyasalam (approved):
lgtm!

AaronCasas (approved):

tessavitabile (approved):

agreenb:
@tessavitabile Since this is a breaking change, would need to be force-merged if possible. Thanks!

agreenb:
@tessavitabile Would it be possible for you to force-merge this today? Confirming that this field is not used in app or RDK.

tessavitabile:
Thanks for pinging me, I missed the last message on this thread.

agreenb:
@tessavitabile Thank you!

PR #146, opened 2023-01-03
[DATA-975] Delete files that were removed from proto.

AaronCasas:
These files were generated when I forgot to put the mltraining proto in a `/v1` folder. After moving it, these files were not removed. This removes them.


agreenb (approved):
Thanks for catching this

PR #148, opened 2023-01-04
APP-1125 adding filter to the robotpartlogs function

RoxyFarhad:
Adding filter to the app.proto GetRobotPartLogsRequest function


michaellee1019 (approved):
As discussed in slack, the generated code will automatically get generated when the PR is approved, so only need to commit the `proto` changes. No big deal since its already committed, no need to remove those files.

RoxyFarhad:
Thanks! Yeah from now on will just push the photo file :)

PR #149, opened 2023-01-04
[DATA-978] Cleanup packages.proto

agreenb:
### Remove `created_on` from `PackageInfo`
After initially being added into the API, this was immediately moved to InternalPackage and the returned Package instead, but this field wasn't removed since it was a breaking change.

The reasoning is that we use PackageInfo when setting the CreatePackageRequest and only want to include it in the GetPackageResponse rather being able to set it in a request. 

Field unused in app/RDK.

### Rename uri->url in `Package`
We're technically always returning a URL.
Field unused in app/RDK.


proto/viam/app/packages/v1/packages.proto
```diff
@@ -57,7 +57,6 @@ message PackageInfo {
   PackageType type = 4;
   repeated FileInfo files = 5;
   google.protobuf.Struct metadata = 6;
-  google.protobuf.Timestamp created_on = 7;
 }
 
 message CreatePackageRequest {
```
tahiyasalam, line 62: I think this relates to more generally thinking about changes with `Packages`, as described [here](https://viam.atlassian.net/browse/DATA-978), and we may want to discuss change uri --> url  cc @AaronCasas

agreenb: Thanks for creating a ticket from that other PR discussion, I realize I made a duplicate! I think `InternalPackage` naming makes sense, and agree on being more precise about URL. Updated.

tessavitabile (approved):
LGTM. Do you need to do anything to coordinate the release of these breaking changes?

agreenb:
Nope! Both this and the other PR are updating APIs that were newly added for internal data usage but never used in app/RDK.

agreenb:
@tessavitabile Would it be possible for you to force-merge this today? Confirming that these fields aren't (and were never) being used anywhere in rdk or app.

agreenb:
@tessavitabile Thank you!

PR #152, opened 2023-01-04
APP-1125 making filter optional

RoxyFarhad:


DTCurrie:
@AdamMagaluk @edaniels The linting check is throwing an error on this but we want to add the `optional` flag to a field that was added to avoid breaking FE checks in `app`. Can we override this lint check?

DTCurrie (approved):
✅

edaniels:
Yes. It's a breaking change but rationale approved.

PR #155, opened 2023-01-04
[RSDK-1060] SLAM Proto Changes

jeremyhyde-viam:
Adding additional endpoints for map representation project (see [scope doc](https://docs.google.com/document/d/13OEt7HX33605dtl-uN1ESNTiB8jpRk-W9oquI16WYGM/edit?usp=sharing)). These endpoints will either replace existing endpoints or add new functionalities to the SLAM system.

- `GetPointCloudMap`: Replaces `GetMap` (`GetMap` will be deleted)
- `GetPositionNew`: Will be renamed to `GetPosition` (current `GetPosition` will be deleted)
- `GetInternalState`: New feature for getting slam algorithm's full map state

JIRA Ticket: [RSDK-1060](https://viam.atlassian.net/browse/RSDK-1060)

[RSDK-1060]: https://viam.atlassian.net/browse/RSDK-1060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-1060]: https://viam.atlassian.net/browse/RSDK-1060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Open Question: is this still v1 or should this be v2 for SLAM


proto/viam/service/slam/v1/slam.proto
```diff
@@ -66,3 +89,42 @@ message GetMapResponse {
...
+  string component_reference = 2;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
+}
+
+message GetPointCloudMapRequest {
+  // Name of slam service
+  string name = 1;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
```
tessavitabile, line 113: Is this needed?

proto/viam/service/slam/v1/slam.proto
```diff
@@ -66,3 +89,42 @@ message GetMapResponse {
   // Actual MIME type of response (image/jpeg or image/pcd)
   string mime_type = 3;
 }
+
+message GetPositionNewRequest {
+  // Name of slam service
+  string name = 1;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
```
tessavitabile, line 97: Do we need `extra` for `GetPosition`? If not, we may be able to exclude it here, since we're already making a breaking change to `GetPosition`.

proto/viam/service/slam/v1/slam.proto
```diff
@@ -26,6 +26,29 @@ service SLAMService {
       get: "/viam/api/v1/service/slam/{name}/map"
     };
   }
+
+  // GetPositionNew returns the current estimated position of the robot with
+  // respect to a returned reference frame
```
tessavitabile, line 31: This should say "component reference" instead of "reference frame" (explicitly excluding the frame system).

proto/viam/service/slam/v1/slam.proto
```diff
@@ -26,6 +26,29 @@ service SLAMService {
...
+  }
+
+  // GetPointCloudMap returns the latest point cloud map generated by the
+  // SLAM library
+  rpc GetPointCloudMap(GetPointCloudMapRequest) returns (GetPointCloudMapResponse) {
+    option (google.api.http) = {
+      get: "/viam/api/v1/service/slam/{name}/point_cloud_map"
+    };
+  }
+  // GetInternalState returns the current map state as defined by the specified
```
tessavitabile, line 45: I would say "the internal state required to continue mapping/localizing".

proto/viam/service/slam/v1/slam.proto
jeremyhyde-viam: In the current proto, the field extra appears in the `GetPositionRequest`, `GetPositionResponse` and `GetMapResponse`. It is currently only being used in the `GetPositionResponse` to return the quaternion (although once the spatial math lib is available in rust, if not already, we will not need it anymore.

Since this field was not discussed in the tech spec I kept them in. However, we can freely remove it from the other two proto messages if desired.

tessavitabile: I think we can remove `extra` from `GetPositionNewRequest` (which will eventually become `GetPositionRequest`), since we are already making a breaking change to `GetPosition`. I think we don't need to include `extra` in `GetPointCloudMapResponse`.

proto/viam/service/slam/v1/slam.proto
jeremyhyde-viam: Changed to `GetInternalState returns the internal map as defined by the specified slam algorithm required to continue mapping/localizing` as I wanted to note that the returned internal state differences based on the algorithm. Let me know if you'd like me to remove the `... as defined by the specified slam algorithm`.

proto/viam/service/slam/v1/slam.proto
jeremyhyde-viam: See comment below

proto/viam/service/slam/v1/slam.proto
```diff
@@ -26,6 +26,29 @@ service SLAMService {
       get: "/viam/api/v1/service/slam/{name}/map"
     };
   }
+
+  // GetPositionNew returns the current estimated position of the robot with
```
edaniels, line 30: Using the word New in GetPositionNew confuses me as to what it would return to me until after reading the comment. What about GetEstimatedPosition?

proto/viam/service/slam/v1/slam.proto
edaniels: can we make a note on this that it's not intended for end user use or something of the like?

proto/viam/service/slam/v1/slam.proto
```diff
@@ -66,3 +89,42 @@ message GetMapResponse {
...
+  google.protobuf.Struct extra = 99;
+}
+
+message GetPositionNewResponse {
+  // Current position of the specified component in the SLAM Map
+  common.v1.Pose pose = 1;
+  // This is usually the name of the camera that is in the SLAM config
+  string component_reference = 2;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
```
edaniels, line 106: extra not needed

edaniels (changes requested):

proto/viam/service/slam/v1/slam.proto
tessavitabile: We won't have GetPositionNew eventually. This is to allow us to do all the development and then swap the functions that get called in the UI. See tech spec: https://docs.google.com/document/d/1GAnAGrv0gcQkIMzY8x3NNanigRzkm6oPH6pKTHg8XjY/edit

edaniels: Maybe a comment explaining that then would help. Otherwise it makes New seem like it's creating something to an unaware user.

jeremyhyde-viam: Adding a comment to note that this function will be replacing the current `GetPosition`

edaniels (approved):

proto/viam/service/slam/v1/slam.proto
```diff
@@ -66,3 +91,38 @@ message GetMapResponse {
...
+  // Name of slam service
+  string name = 1;
+}
+
+message GetPositionNewResponse {
+  // Current position of the specified component in the SLAM Map
+  common.v1.Pose pose = 1;
+  // This is usually the name of the camera that is in the SLAM config
+  string component_reference = 2;
+  // Additional arguments to the method
```
tessavitabile, line 105: [super nit] Maybe "Additional information in the response"

tessavitabile (approved):

PR #158, opened 2023-01-05
APP 1144: Adding pagination to GetRobotPartLogs

RoxyFarhad:


proto/viam/app/v1/app.proto
```diff
@@ -269,11 +275,13 @@ message LogEntry {
 
 message GetRobotPartLogsResponse {
   repeated LogEntry logs = 1;
+  optional pageToken next_page_token = 5;
 }
 
 message TailRobotPartLogsRequest {
   string id = 1;
   bool errors_only = 2;
+  optional string filter = 3;
```
RoxyFarhad, line 284: not sure if we actually want this -- should we make it s.t. this endpoint only ever returns live functions without a filter? Thinking about it more that makes more sense to me

proto/viam/app/v1/app.proto
```diff
@@ -250,10 +250,16 @@ message GetRobotPartResponse {
...
+message PageToken {
+  string log_id = 1;
+}
+
 message GetRobotPartLogsRequest {
   string id = 1;
   bool errors_only = 2;
   optional string filter = 3;
+  optional int64 offset = 4;
+  optional PageToken page_token = 5;
```
AdamMagaluk, line 262: It can just be a string type instead of PageToken message.

proto/viam/app/v1/app.proto
```diff
@@ -250,10 +250,16 @@ message GetRobotPartResponse {
...
 
+message PageToken {
+  string log_id = 1;
+}
+
 message GetRobotPartLogsRequest {
   string id = 1;
   bool errors_only = 2;
   optional string filter = 3;
+  optional int64 offset = 4;
```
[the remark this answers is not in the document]
RoxyFarhad: ah yeah that was a mistake to leave that in there

proto/viam/app/v1/app.proto
```diff
@@ -269,11 +275,13 @@ message LogEntry {
 
 message GetRobotPartLogsResponse {
   repeated LogEntry logs = 1;
+  optional PageToken next_page_token = 5;
```
AdamMagaluk, line 278: PageToken -> string

AdamMagaluk: And remove the optional.

RoxyFarhad: need the optional to not break the frontend for backwards compatibility -- ran into this yesterday with the filter

proto/viam/app/v1/app.proto
AdamMagaluk: The reason to encode this is because the token should be opaque to the client. The client does not care and show not know if we use the ObjectID or anything else. It just needs a token which is a string to be passed back and forth.

proto/viam/app/v1/app.proto
AdamMagaluk: hmm. Do we construct this on the FE anywhere. It should only complain if we are creating a object and don't supply all the fields. In this case it shouldn't matter.

We need to be able to support adding new fields without optional or it's going to make our apis really cumbersome to use. proto3 allows new fields to be added without marking them as optional.

RoxyFarhad: Hmm actually maybe it's something that is bound to Request objects with the linter. It would make sense that the linter doesn't complain to me for the response without explicating setting optional but let me try locally.

RoxyFarhad: yup this is fine w/o the optional here too

proto/viam/app/v1/app.proto
AdamMagaluk: I think it makes sense to have it. We will want a filter with tail eventually.

PR #159, opened 2023-01-06
APP-1125: Make filter a struct rather than a string

RoxyFarhad:


AdamMagaluk:
There was some misunderstanding. I think leaving the proto with just a string field is fine for now. No need for a message structure. We can add it later if needed. But for the backend impl i wanted to only do the refactoring once since it involves changing lots of files in data/ and auth/ directories.

RoxyFarhad:
okay -- ill leave it as is, and keep the filter as a string

PR #160, opened 2023-01-06
APP-1146: Publish api buf artifacts with named tag matching git tag

AdamMagaluk:
Fork code changes: https://github.com/bufbuild/buf-push-action/compare/main...AdamMagaluk:buf-push-action:main

I don't have 100% confidence in this and there isn't a great way to test this without pushing FYI.

### DIFF
--- a/.github/workflows/tag_and_release_protos.yml
+++ b/.github/workflows/tag_and_release_protos.yml
@@ -22,9 +22,12 @@ jobs:
     container:
       image: ghcr.io/viamrobotics/canon:amd64-cache
       options: --platform linux/amd64
+    outputs:
+      new_tag: ${{ steps.bump.outputs.new_tag }}
     steps:
       - uses: actions/checkout@v3
       - name: Bump tag
+        id: bump
         uses: anothrNick/github-tag-action@1.52.0
         env:
           GITHUB_TOKEN: ${{ secrets.REPO_READ_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tag_and_release_protos.yml
```diff
@@ -22,27 +22,32 @@ jobs:
...
       - name: Bump tag
+        id: bump
         uses: anothrNick/github-tag-action@1.52.0
         env:
           GITHUB_TOKEN: ${{ secrets.REPO_READ_TOKEN }}
           WITH_V: true
           DEFAULT_BUMP: patch
 
   buf-push:
+    needs: bump-tags
```
michaellee1019, line 38: Should be `bump-tag`

michaellee1019 (commented):
Nice! you did end up forking it :) I thinking i found a bug. Otherwise worth a shot and LGTM

proto/viam/app/v1/app.proto
```diff
@@ -161,7 +161,6 @@ message RobotPartHistoryEntry {
   RobotPart old = 4 [(tagger.v1.tags) = "bson:\"old\" json:\"old\""];
 }
 
-// Organizations
```
edaniels, line 164: undo?

edaniels (approved):
If it works it works, lgtm

proto/viam/app/v1/app.proto
AdamMagaluk: I need a file in proto/** change for it to kick off the workflow. And this wasn't a great comment to begin with.

.github/workflows/tag_and_release_protos.yml
AdamMagaluk: Good catch thanks.

edaniels (approved):
LGTM

PR #162, opened 2023-01-09
Remove Deprecated FindRobots

michaellee1019:
FindRobots was renamed to ListRobots. APP+RDK was updated to point to the ListRobots.

This will be a breaking change from proto perspective. However all usages of FindRobots has been replaced with ListRobots, so no actual breaking change exists.


proto/viam/app/v1/app.proto
```diff
@@ -111,9 +111,6 @@ service AppService {
   // Get a list of robots
   rpc ListRobots(ListRobotsRequest) returns (ListRobotsResponse);
 
-  // Finds robots given a query. Deprecated, use ListRobots instead
-  rpc FindRobots(FindRobotsRequest) returns (FindRobotsResponse);
```
AdamMagaluk, line 115: I would typically say its best practice to remove all implementation usage before deleting the proto. This ensures that someone else bumping the proto version doesn't need to worry about fixing the breaking changes. See: https://cs.github.com/viamrobotics/app/blob/c6f04c7f8e699cff8d94a9cc59ef764d265a2ae5/service/server.go?q=FindRobotsRequest#L1094

Also it helps double checking if there is usage. Eg. searching across github that there is no references to the code. (outside of generated code)

AdamMagaluk (approved):

proto/viam/app/v1/app.proto
michaellee1019: Good point about doing a company wide search. I swear I removed the code in the App removal PR but looks like I forgot. I'll do that now :)

PR #167, opened 2023-01-23
DATA-815 - Define Packages config

AdamMagaluk:
Defines the gRPC and proto Packages config.


proto/viam/app/v1/robot.proto
```diff
@@ -320,3 +321,13 @@ message ModuleConfig {
   // path to the executable
   string path = 2;
 }
+
+// PackageConfig is the configration for deployed Packages.
+message PackageConfig {
+  // Name is the local name of the package on the RDK. Must be unique across Packages. Must not be empty.
+  string name = 1;
+  // Package is the unqiue package name hosted by Viam. Must not be empty.
```
agreenb, line 329: [nit] Helpful to specify that the format is `<organization_id>/<viam_package_name>`
[supernit] spelling of unique

To expand on the conversation in the scope doc: with shared locations, could download two Viam packages with the same name but under different orgs to the same robot — this would work since the local `name` would be different.

agreenb (approved):
LGTM, thank you for adding so quickly.

proto/viam/app/v1/robot.proto
AdamMagaluk: I'd suggest to leave out the format. This is should be unknown to the users and they should never have to manually create this package string. Our robot config would automatically be generated by the app who controls the naming and the RDK should just pass it along to the package manager and ultimately the RPC request to down.

When you document the structure of the ID you lock yourself in to not being able to change it later and make false assumptions on client(s). 

Yes but the Package format should also be globally unique within Viam with organization_id/viam_package_name. Where as the `name` is locally unique to the robot.

agreenb (approved):
LGTM

proto/viam/app/v1/robot.proto
agreenb: Since app always controls the naming of packages, that's a good point that this should always be unknown to users. Agreed.

agreenb (approved):

PR #170, opened 2023-01-23
APP-1216: Create Fragment Request & Response types for new service calls

RoxyFarhad:


proto/viam/app/v1/app.proto
```diff
@@ -508,8 +508,34 @@ message Fragment {
   string organization_owner = 4 [(tagger.v1.tags) = "bson:\"organization_owner\" json:\"owner\""];
   bool public = 5 [(tagger.v1.tags) = "bson:\"public\" json:\"public\""];
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
+  string organization_name = 7;
```
RoxyFarhad, line 511: We don't have this in the db Fragment model, but we return it to the FE to be used so it should be in the definition.

proto/viam/app/v1/app.proto
```diff
@@ -508,8 +508,34 @@ message Fragment {
...
 }
 
+message FragmentCreateRequest {
```
RoxyFarhad, line 514: These requests match the format of our current Fragment json requests/responses

proto/viam/app/v1/app.proto
```diff
@@ -119,6 +119,23 @@ service AppService {
 
   // DeleteRobot deletes a robot
   rpc DeleteRobot(DeleteRobotRequest) returns (DeleteRobotResponse);
+
+  // Fragments
+
+  // Gets a list of fragments
+  rpc ListFragments(ListFragmentsRequest) returns (ListFragmentsResponse);
```
RoxyFarhad, line 126: The new requests based on the app's fragments.go and fragment.go pages

proto/viam/app/v1/app.proto
```diff
@@ -508,8 +525,48 @@ message Fragment {
...
+  string id = 1;
+}
+
+message GetFragmentResponse {
+  Fragment fragment = 1;
+}
+
+message CreateFragmentRequest {
+  string name = 1;
+  google.protobuf.Struct fragment = 2;
```
AdamMagaluk, line 547: Maybe rename this to config to be in parity with robot part?

proto/viam/app/v1/app.proto
```diff
@@ -508,8 +525,48 @@ message Fragment {
   string organization_owner = 4 [(tagger.v1.tags) = "bson:\"organization_owner\" json:\"owner\""];
   bool public = 5 [(tagger.v1.tags) = "bson:\"public\" json:\"public\""];
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
+  string organization_name = 7;
 }
 
+message ListFragmentsRequest {}
```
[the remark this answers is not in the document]
RoxyFarhad: From my understanding the current http request:

```
func (p *Fragments) ServeAPI(w http.ResponseWriter, r *http.Request) (interface{}, error) {
	ctx := r.Context()
	state := auth.MustContextState(ctx)
	if !state.LoggedIn() {
		return nil, errors.New("have to be logged in")
	}

	var err error
	
	Fragments := []*fragmentTableData{}

	var fragments []*data.Fragment

	fragments, err = p.DB.FragmentFindAll(ctx)
	if err != nil {
		return nil, err
	}

	for _, f := range fragments {
		fragmentStruct, err := structpb.NewStruct(f.Fragment)
		if err != nil {
			return nil, err
		}

		fragment := &fragmentTableData{
			ID:               f.ID,
			Name:             f.Name,
			Public:           f.Public,
			OrganizationName: "",
			OrganizationID:   "",
			Fragment:         fragmentStruct,
		}

		for _, o := range state.MyOrgs {
			if f.OrganizationOwnerID == o.ID {
				fragment.OrganizationName = o.Name
				fragment.OrganizationID = o.ID
			}
		}

		Fragments = append(Fragments, fragment)
	}

	return Fragments, nil
}
```

returns all the fragments on the server? but we do seem to hide org names / ids for fragments outside the org? This will probably need to change down the line, but I am just trying to maintain default behavior right now

proto/viam/app/v1/app.proto
```diff
@@ -508,8 +525,49 @@ message Fragment {
   string organization_owner = 4 [(tagger.v1.tags) = "bson:\"organization_owner\" json:\"owner\""];
   bool public = 5 [(tagger.v1.tags) = "bson:\"public\" json:\"public\""];
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
+  string organization_name = 7;
+  bool readonly = 8;
```
RoxyFarhad, line 529: For the getFragment request, we return this field, even though it's not currently in the proto definition. It is currently always set to false, but adding this in here just to mimic current behavior.

proto/viam/app/v1/app.proto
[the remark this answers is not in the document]
AdamMagaluk: It's a bit convoluted but our permissioned data implementation filters out only fragments the user can read on the request baed on the users state for the current org.

But I would not try to maintain the implementation here but instead maintain only what's needed to build our product. We display all public fragments and org fragments when you view them. So we need to just support that.

With the API we do not have user state so we cannot infer which fragment the user can access or not without returning all fragments across all orgs they have access to which breaks the current behavior of the app.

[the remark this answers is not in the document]
RoxyFarhad: yup, that's true that FindAll fragments will only return what the user has access to based on org etc. But confused what you mean here in the context of this request, and how the request params should look?

AdamMagaluk (approved):
LGTM thanks!

AdamMagaluk (approved):

PR #171, opened 2023-01-24
[DATA-1054] Add checksum to packages

agreenb:
Add checksum to packages so that RDK does not need to redownload existing packages.


proto/viam/app/packages/v1/packages.proto
```diff
@@ -81,6 +81,7 @@ message Package {
   PackageInfo info = 1;
   string url = 2;
   google.protobuf.Timestamp created_on = 3;
+  string checksum = 4;
```
AdamMagaluk, line 84: FYI: I think this should match the GCS CRC32C that will be available. See: https://cloud.google.com/storage/docs/hashes-etags

proto/viam/app/packages/v1/packages.proto
```diff
@@ -93,6 +94,7 @@ message InternalPackage {
   google.protobuf.Struct metadata = 6 [(tagger.v1.tags) = "bson:\"metadata\" json:\"metadata\""];
   string blob_path = 7 [(tagger.v1.tags) = "bson:\"blob_path\" json:\"blob_path\""];
   google.protobuf.Timestamp created_on = 8 [(tagger.v1.tags) = "bson:\"created_on\" json:\"created_on\""];
+  string checksum = 9 [(tagger.v1.tags) = "bson:\"checksum\" json:\"checksum\""];
```
AdamMagaluk, line 97: [nit] For the future we shouldn't add `InternalPackage` to our API proto if it never is surfaced in our APIs. Instead we should define golang struct DB models in the app that are used for mongo vs our APIs. The fleet team is migrating all our usage to that model now.

proto/viam/app/packages/v1/packages.proto
```diff
@@ -103,6 +105,7 @@ message GetPackageRequest {
 
 message GetPackageResponse {
   Package package = 1;
+  string version = 2;
```
AdamMagaluk, line 108: Actually this may not be needed. I didn't notice but PackageInfo has `string version = 3; `. I think that would suffice.

AdamMagaluk (approved):

proto/viam/app/packages/v1/packages.proto
agreenb: We were considering what the best practice was was initially defining models that are used for Mongo, so this is great to know. Created https://viam.atlassian.net/browse/DATA-1057 to track in our backlog.

proto/viam/app/packages/v1/packages.proto
agreenb: You're right, removed.

AdamMagaluk (approved):
LGTM thanks

PR #175, opened 2023-01-25
remove ExportPointCloud

nfranczak:
this PR removes ExportPointCloud from the motion service.
This is a breaking change.


biotinker (approved):
LGTM.

This may require manual merge from either @edaniels or @Otterverse since it's technically a "breaking change" even though it's only been around since this morning

PR #176, opened 2023-01-25
[DATA-1014]/[DATA-1015] Add processing bool to TrainingJob.

AaronCasas:
Accompanying app [PR](https://github.com/viamrobotics/app/pull/1108).

Removes `Submitting` state, which was used to signify Pending jobs that had been picked up but not yet finished submitting to Vertex, with a `processing` bool, which more generally indicate that a particular Job is being processed by some JobManager. This was done because I found that we were going to need a similar intermediary state between InProgress and Completed, so figured it made it a little more future proof to add a more general concept for enforcing mutual exclusion than adding another intermediary state.


agreenb:
@AaronCasas Since this is an internal-only DB representation and we're going to move that and other similar instances (e.g. InternalPackage) over to Golang structs like Fleet is in the process of doing, I'd just made this an additive non-breaking change of the `processing` bool field. 

Then can do a fast-follow of moving to a struct, since TrainingJob is only used within db.go and ml_training_service.go

AaronCasas:
@agreenb makes sense, removed. This does still keep the breaking change re removing the Submitting status, though

agreenb (approved):
LGTM, and confirming that the SUBMITTING status is currently only set in db.go, which is removed in the corresponding app PR.

agreenb:
@tessavitabile Could you force-merge this? The breaking change is removing an enum SUBMITTING status for the in-development ML Training Service. The only use of it was in db.go, which is removed in the corresponding app PR. This isn't being used in the FE or anywhere else yet.

@AaronCasas [supernit] Could you update the PR description to remove the reference to ModelMetadata?

agreenb:
Thanks @tessavitabile!

PR #179, opened 2023-01-30
APP-1231: Adding use_count to fragment pb model

RoxyFarhad:
We want to show usage statistics for a fragment, so this allows us to return that.


proto/viam/app/v1/app.proto
```diff
@@ -527,6 +527,7 @@ message Fragment {
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
   string organization_name = 7;
   bool readonly = 8;
+  int32 use_count = 9;
```
AdamMagaluk, line 530: Not sure if it will be used for this but we should avoid relying on use_count to signify if the frontend should be allowed to delete the fragment. Use a different field for that like `delete_allowed`.

AdamMagaluk (approved):

proto/viam/app/v1/app.proto
RoxyFarhad: we want to add a column showing the number of robots using this fragment, which is what the number signifies. Delete allowed isn't visible, and currently just blocks the delete if you try to delete a fragment that is being used somewhere.

PR #180, opened 2023-01-31
APP-1126: Adding robot & org use count

RoxyFarhad:
Adding org use count to the fragment and renaming use_count to robot use count for clarity


RoxyFarhad:
@stevebriskin I was wondering if you can force merge this, use_count is not currently being used anywhere and this makes the distinction clearer between whether its robot_part use_count or organization use_count. When I added the orgiinal field use_count, I missed that we would also need a count for organizations.

proto/viam/app/v1/app.proto
```diff
@@ -527,7 +527,8 @@ message Fragment {
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
   string organization_name = 7;
   bool readonly = 8;
-  int32 use_count = 9;
+  int32 robot_part_count = 9;
+  int32 organization_count = 10;
```
AdamMagaluk, line 531: Can you add comment about what this count means? Same with the above.

michaellee1019 (commented):
Just discussed in person. Field comments should be placed on line above field. Otherwise looks good!

michaellee1019 (approved):

PR #181, opened 2023-02-01
[DATA-1053] Add package ID

agreenb:
Adds package `id` to replace a combined <organization_id, name> identifier. This keeps the package identifier opaque to the client, flexible to change on the server-side, and easy to pass around between Create and Get.

The id is returned in `CreatePackage` and populated in the `Package` message retrieved by `GetPackage` and `ListPackages` RPCs, and can be used to get/delete the package in `GetPackageRequest`/`DeletePackageRequest`.

The plan is still to store a package by <organization_id, name, version> in the internal DB representation, but this allows us to migrate that later.

Note: 
* Will follow-up to remove `organization_id` and `name` from GetPackageRequest.
* Not including it in the ListPackagesRequest since that one is intended as a way to filter for multiple packages.


proto/viam/app/packages/v1/packages.proto
```diff
@@ -102,6 +105,7 @@ message GetPackageRequest {
   string organization_id = 1;
   string name = 2;
   string version = 3;
+  string id = 4;
```
AdamMagaluk, line 108: Docs for this. And what the desired logic is for when org and name are supplied with id?  ^ same on delete.

AdamMagaluk: Ah sorry completely missed that and went right to the Files tab :(

AdamMagaluk (approved):

proto/viam/app/packages/v1/packages.proto
agreenb: Note that just added `version` in the CreatePackageResponse to make it return everything you need to call Get/Delete.

AdamMagaluk: Should we also add the filed to sepcify if we should generated a signed URL? Can be in a follow up.

PR #183, opened 2023-02-01
APP-1127: Adding optional param public to change from public->private…

RoxyFarhad:
… or vice-versa


proto/viam/app/v1/app.proto
```diff
@@ -563,6 +563,7 @@ message UpdateFragmentRequest {
   string id = 1;
   string name = 2;
   google.protobuf.Struct config = 3;
+  optional bool public = 4;
```
RoxyFarhad, line 566: Optional for now to make sure nothing on the FE breaks (because we don't currently pass it in)

AdamMagaluk (approved):

proto/viam/app/v1/app.proto
RoxyFarhad: We can / will , but the issue is if this isn't optional the time between this PR merges and the APP merges this request will break.

DTCurrie (approved):
✅

PR #184, opened 2023-02-01
APP-1252 Add robot_count field to Location message

DTCurrie:
Adding a `robot_count` field to the `Location` message, since it is a common use case to want to know the amount of robots, if any, in a location.


michaellee1019 (approved):

AdamMagaluk (approved):

DTCurrie:
Reopeneing

PR #189, opened 2023-02-06
APP-1259: Indicate whether the fragment is used outside of its owner org

RoxyFarhad:
This is to hide the warning message for changing the fragment from public -> private if the fragment is only used by the owner org. (The bug that you found out last week)


proto/viam/app/v1/app.proto
```diff
@@ -541,6 +541,8 @@ message Fragment {
   int32 robot_part_count = 9;
   // number of organizations using this fragment
   int32 organization_count = 10;
+  // whether the organization(s) using this fragment is the same as the fragment org
+  bool is_only_used_by_owner = 11;
```
michaellee1019, line 545: To match convention, I would change to `only_used_by_owner`. The rest of the boolean fields do not have `is_` prefix.

michaellee1019 (approved):
I have a thought. If we returned a list of org ids, that would also then

proto/viam/app/v1/app.proto
RoxyFarhad: Thanks sounds good!

RoxyFarhad:
@michaellee1019 I'm refactoring the BE to use the organization count and onlyUsedByOwnerOrg to be in the same query, if that matches what you think your thought is?

michaellee1019:
> @michaellee1019 I'm refactoring the BE to use the organization count and onlyUsedByOwnerOrg to be in the same query, if that matches what you think your thought is?

Makes sense. If thats the case returning 1 as total and true as onlyUsedByOwnerOrg. Now that you bring it up... Another option could be that we return a list of orgIds instead of these two properties. I think that would cover what we need? If we have the list of ids we can make a check of size and `ids[0]=owner` org?

PR #191, opened 2023-02-08
APP-1288: GetRoverRentalRobots

RoxyFarhad:
For the Rover Reuse V2, we want to be able to have the option to re-use a robot for the rental. To do so, we need an endpoint that will return all robots in the "Rover Rental" location if it exists. We want  to show the Robot Name, and require the Robot ID to create a reservation so that is what we return from the request.


proto/viam/app/v1/app.proto
```diff
@@ -441,6 +444,19 @@ message GetRobotRequest {
   string id = 1;
 }
 
+message GetRoverRentalRobotsRequest {}
```
RoxyFarhad, line 447: We don;t pass in anything because its harcoded to be robots in Location with name "Rover Rental". Checking on a string is bad practice, but this is the plan for now.

proto/viam/app/v1/app.proto
```diff
@@ -441,6 +444,18 @@ message GetRobotRequest {
...
+
+message RoverRentalRobot {
+  string robot_id = 1;
```
RoxyFarhad, line 450: we only need the robot_id because we will by default create a reservation with the "main" part of the robot.

proto/viam/app/v1/app.proto
```diff
@@ -441,6 +444,18 @@ message GetRobotRequest {
...
+  string location_id = 3;
```
anjinai, line 451: Should this be = 2 ?

proto/viam/app/v1/app.proto
```diff
@@ -441,6 +444,18 @@ message GetRobotRequest {
...
+  string robot_name = 4;
```
anjinai, line 452: Should this be = 3 ?

RoxyFarhad: ah yeah, removed a field and forgot to update the numbers :/ re-pushed

anjinai (approved):
LGTM - except for those two questions I left

PR #193, opened 2023-02-13
[DATA-1110] Add ListTrainingJobs rpc.

AaronCasas:
Ticket: https://viam.atlassian.net/browse/DATA-1110


proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -42,6 +43,15 @@ message GetTrainingJobResponse {
   TrainingJobMetadata metadata = 2;
 }
 
+message ListTrainingJobsRequest {
+  string organization_id = 1;
+  TrainingStatus status = 2;
+}
+
+message ListTrainingJobsResponse {
+  repeated TrainingJobMetadata jobs = 1;
```
alexis-wei, line 52: Unless `synced_model_id ` and `id` are the same value, I think we will also need the `id` of these training jobs. For things like cancelling and deletion in the future. So would make a new message type (maybe `TrainingJobInfo`?) that contains the `id` and the `metadata` and return a repeated of that.

alexis-wei: can optionally also replace the return type of `GetTrainingJobResponse` to that in addition

alexis-wei (changes requested):

proto/viam/app/mltraining/v1/ml_training.proto
agreenb: Good callout. It's a refactor, but it seems like an improvement just move id into TrainingJobMetadata? And have the TrainingJob DB _id use the nested TrainingJobMetadata.id?

I don't want this to be a blocker right now so I'd also be ok with adding repeated ids here to make it directly 1:1 map to GetTrainingJobResponse, then shift both over as a cleanup.

AaronCasas: So in the relatively near term TrainingJob shouldn't even exist in api - I think it would be the least work to just yeet that now, move the _id into TrainingJobMetadata (and probably rename to TrainingJob), then make the current TrainingJob just a go struct in app

agreenb: Sounds good, if it's easiest/possible to do so within the current app PR or a separate, happy to review that pretty quickly. I'd make this a nonbreaking change in API and then cleanup after.

AaronCasas: Changed  to put id in TrainingJobMetadata, and stop using TrainingJob entirely, and added a `dbTrainingJob` struct fulfilling the same purpose TrainingJob was

AaronCasas: Now data/mltraining, GetTrainingJobResponse.id, and TrainingJob are all deprecated/unused. I think we should avoid future changes after this  until we address/remove all of those.

agreenb: Thank you! I updated https://viam.atlassian.net/browse/DATA-1071 from being specific to removing TrainingJob to all of these API changes. Moved it from our current sprint to the next one, since it's a cleanup whereas we're focusing on feature completion and testing the training service this week.

alexis-wei (approved):
LOOKS GOOD!

PR #198, opened 2023-02-16
RSDK-1942 Add DoCommand to every component proto

benjirewis:
RSDK-1942

Adds a new `DoCommand` rpc method on all relevant component protos and complies protos.


proto/viam/buf.yaml
```diff
@@ -18,6 +18,5 @@ lint:
       - component/camera/v1/camera.proto
       - component/audioinput/v1/audioinput.proto
     RPC_REQUEST_RESPONSE_UNIQUE:
-      - component/camera/v1/camera.proto
-      - component/audioinput/v1/audioinput.proto
+      - component
```
benjirewis, line 21: I've ignored (for all of the `component` directory) the default [`buf lint` requirement](https://docs.buf.build/lint/configuration#rpc_allow_) that all rpc methods have unique request and response message types since it defeats the point of this project. We'll have to ignore `service` as well, @purplenicole730.

cheukt (commented):
this generally looks good - can you actually point your rdk to these changes with `go mod edit -replace go.viam.com/api=../<local path to your api repo>` and try running the components? just want to make sure things aren't broken and they shouldn't be

purplenicole730 (approved):

proto/viam/buf.yaml
purplenicole730: How do we ignore the requirements while running `make dist/buf`?

benjirewis: As discussed offline you'll have to add `- service` here, too. That should remove the warning for both linting and compilation.

benjirewis:
@cheukt `go test -v count=1 ./components/*` passes, and running curls like `grpcurl  -plaintext  -d='{"name": "base", "command": {"cmd": "test", "data1": 500}}' localhost:8080 viam.component.base.v1.BaseService/DoCommand` go through but return unimplemented errors.

cheukt (approved):

PR #204, opened 2023-02-24
APP-1305: Adding orgId to the rover rental request

RoxyFarhad:
I thought that we could get a user's org based on the Context state in the Server to get the Rover Rental Location for that Organization. This is actually incorrect, and so we need to pass it in with the GetRoverRentalRobots Request.


ehhong (approved):
lgtm

PR #206, opened 2023-02-27
[DATA-1010/DATA-1071]: Remove ML training deprecated API

tahiyasalam:
In this PR:
- Remove `TrainingJob` from the ML training APIs
- Remove `GetTrainingJob.id`
- Remove MLTraining RPCs from DataService

This also requires changes across RDK [here](https://github.com/viamrobotics/rdk/pull/1941) and APP [here](https://github.com/viamrobotics/app/pull/1236). 

Once I get approval from other folks, I will add Tess for approval/force merge. Otherwise, they will need to re-approve after each requested change.


proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -69,14 +67,3 @@ message TrainingJobMetadata {
   string user_email = 6 [(tagger.v1.tags) = "bson:\"user_email\" json:\"user_email\""];
   string id = 7 [(tagger.v1.tags) = "bson:\"_id\" json:\"id,omitempty\""];
 }
-
-message TrainingJob {
```
dmhilly, line 73: Why are we eliminating this? We just use a Go struct instead?

dmhilly (commented):
One question!

proto/viam/app/mltraining/v1/ml_training.proto
tahiyasalam: Yup! Instead, we have a `dbTrainingJob` go struct that we insert into the `jobs` db. 

In this ticket [here](https://viam.atlassian.net/jira/software/c/projects/DATA/boards/40?modal=detail&selectedIssue=DATA-1071), Aaron mentions this is only used internally so no need to exposed in API.

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -39,7 +38,6 @@ message GetTrainingJobRequest {
 }
 
 message GetTrainingJobResponse {
-  string id = 1;
   TrainingJobMetadata metadata = 2;
```
alexis-wei, line 41: change this number to 1

alexis-wei (changes requested):

proto/viam/app/mltraining/v1/ml_training.proto
dmhilly: This can actually cause issues if for example there are robots or servers on app running older versions of rdk, since they will expect gRPC responses where metadata corresponds to index 2, but app will send a message with metadata as field 1. I don't know how or how often rdk updates on robots but I assume they don't all update simultaneously. So we might want to change this at a later point or design this for backwards compatibility. How we handle this depends on which of robots/users/app frontend actually send these requests and whether or not we can be in a state where the backend is running the latest version of app but clients are sending gRPC requests with older versions of rdk or app frontend.

AaronCasas: In general I'd totally agree with the motivation to maintain backwards compatibility, but in this case I think:

- The only thing using this is app, and we control the version + deployment there.
- No other clients are using this, so there isn't a risk of a version mismatch.

In this case I think the risk of the breaking change is less than the benefit of keeping the proto as clean as possible for as long as possible

AaronCasas (approved):
🚀

alexis-wei (approved):
LOOKS GOOD!!

PR #217, opened 2023-03-12
Add explicit external auth config for robots

edaniels:
This supports https://github.com/viamrobotics/goutils/pull/145


michaellee1019:
LGTM. Just so I'm clear on why this change is being done, writing the rationale out. Let me know if any of this is wrong:
1. External is a better name for our clients to understand than `WebOAuth`.
2. We are moving a way from the "IsA AuthHandlerConfig" and CredentialType enum because its simpler for users to instantiate a client with a single `jwks` parameter. They don't need to know the different auth types and having to enter `CREDENTIALS_TYPE_WEB_OAUTH` is redundant any annoying for external API callers.
3. Internally we still need many different types of authentication. In some cases we want to specify multiple authentication methods, so we need to keep the field `repeated AuthHandlerConfig handlers = 1;`

edaniels:
> LGTM. Just so I'm clear on why this change is being done, writing the rationale out. Let me know if any of this is wrong:
> 
> 1. External is a better name for our clients to understand than `WebOAuth`.

Yep! It reflects what's happening and is consistent with the original intent in goutils.

> 2. We are moving a way from the "IsA AuthHandlerConfig" and CredentialType enum because its simpler for users to instantiate a client with a single `jwks` parameter. They don't need to know the different auth types and having to enter `CREDENTIALS_TYPE_WEB_OAUTH` is redundant any annoying for external API callers.

For app managed robots, a user won't ever interact with this directly (same as it is now). I don't see this as a user configurable thing as far as our config goes so separating it makes it more clear it's special (fairly subjective).

> 3. Internally we still need many different types of authentication. In some cases we want to specify multiple authentication methods, so we need to keep the field `repeated AuthHandlerConfig handlers = 1;`

Yep! We actually may want multiples of the same type in the future but that's a different story.

michaellee1019:
> I don't see this as a user configurable thing as far as our config goes

Good point!

michaellee1019 (approved):

PR #220, opened 2023-03-23
APP-1218: Adding health service

RoxyFarhad:


proto/viam/health/health.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package viam.health.v1;
+
+option go_package = "go.viam.com/api/health/v1";
+
+enum Service {
+  SERVICE_ALL_UNSPECIFIED = 0;
```
RoxyFarhad, line 8: All services need to be defined here so its easy to do different service checks. 
An enum made sense rather than string values because its easy to switch on those.

PR #232, opened 2023-04-03
Adding AddRole and RemoveRole functions

piokasar:
These need to be included in addition to the AddRoleRequest, RemoveRoleRequest, and ListAuthoriztionRequest pbs.


michaellee1019 (approved):
My fault for not catching earlier :)

piokasar:
> My fault for not catching earlier :)

no worries

PR #233, opened 2023-04-03
APP-1536 Add protos for app billing service that will interface with the Stripe API

ehhong:
**Notes**:
- Stripe can take as input way more information than the request objects take here -- I've included the bare minimum
- Similarly, Stripe returns way more information than response objects return here

We can update these as necessary, we will probably need to do so as we work through all possible billing flows and levels of customization

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -146,6 +146,17 @@ service AppService {
 
   // Deletes a fragment
   rpc DeleteFragment(DeleteFragmentRequest) returns (DeleteFragmentResponse);
+
+  // Billing
+
+  // Creates a Stripe customer
+  rpc CreateStripeCustomer(CreateStripeCustomerRequest) returns (CreateStripeCustomerResponse);
+
+  // Create a Stripe setup intent
+  rpc CreateStripeSetupIntent(CreateStripeSetupIntentRequest) returns (CreateStripeSetupIntentResponse);
+
+  // Create a Stripe payment intent
+  rpc CreateStripePaymentIntent(CreateStripePaymentIntentRequest) returns (CreateStripePaymentIntentResponse);
 }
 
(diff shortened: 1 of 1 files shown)


DTCurrie (approved):
✅

proto/viam/app/v1/app.proto
```diff
@@ -146,6 +146,17 @@ service AppService {
 
   // Deletes a fragment
   rpc DeleteFragment(DeleteFragmentRequest) returns (DeleteFragmentResponse);
+
+  // Billing
+
+  // Creates a Stripe customer
+  rpc CreateStripeCustomer(CreateStripeCustomerRequest) returns (CreateStripeCustomerResponse);
```
RoxyFarhad, line 153: Hm do we need an RPC for this? My assumption was that we can embed this within our current CreateOrganization grpc call

DTCurrie:
@edaniels brought up a good point that any third-party specific APIs should not live in the public API repo. These kind of implementations should probably just exist in the app. Any of the non-stripe billing APIs should live here.

edaniels:
I'd suggest keeping this in a private proto namespace in our API (either in this repo or in app). That way, people know this isn't part of the standard usage of the API and is very much cloud implementation specific.

RoxyFarhad:
I agree with this ^ We should create a new file in APP with protos in it for this integration purpose. My only concern is adding the right tooling so the bufs generate correctly, but copying the commands from rover-rental or api to generate buf files should be fine.

proto/viam/app/v1/app.proto
RoxyFarhad: yeah, unless there is a specific reason for this to exist as something the FE must call don't think we need this.

ehhong:
decided to move these protos into app -- closing this PR

PR #234, opened 2023-04-04
Add additional fields to permissions protos

piokasar:
Main changes:
- Added organization_id field to ListIdentityAuthorizations proto (this is necessary to avoid db calls in canAccessOrg checks in permissioned_data.go in permissions endpoints [here](https://github.com/viamrobotics/app/pull/1381/files#diff-e210d856e0f6f7c8bbaa54fc23bd4808fd1c84ce4efc6837cfe0f9005682b431:~:text=func%20(pd%20*permissionedData)%20ListIdentityAuthorizations,Write))
- Added organization_id field to RemoveRoleRequest proto
- Changes in permissions tech spec protos


piokasar:
@michaellee1019 the lint check is failing due to the breaking changes. The permissions protos shouldn't be a problem I'll just have to make some adjustments to the endpoints.

michaellee1019:
> @michaellee1019 the lint check is failing due to the breaking changes. The permissions protos shouldn't be a problem I'll just have to make some adjustments to the endpoints. 

All good, breaking changes are expected here. It doesnt actually break anything because its not used yet. We'll have steve merge it in for us to override.

michaellee1019:
@stevebriskin can you merge in this breaking change PR? It not actually breaking, includes some changes to permissions protos that were started before tech spec approval, with the latest approved tech spec.

PR #235, opened 2023-04-04
RSDK-2563 Add MoveOnMap endpoint to motion service

raybjork:
This pull request updates the motion service to reflect recently scoped changes to the 2D planning.  `MoveOnMap` will be the endpoint for moving a 2D base with a slam map, while the more general `Move` will handle the rest of the cases


spieswl (approved):
LGTM!. Question for @edaniels: do we still need to commit the auto-generated changes to other go, js, ts files or is that something that can be handled by a workflow?

edaniels:
workflow handles this upon approval

edaniels:
with the protos-compiled label

PR #236, opened 2023-04-04
APP-1526: Adding billing proto methods

RoxyFarhad:
These methods are the barebones of what we need for the Frontend to show a Billing Page. I put these in public api rather than app/billing proto file as in theory we may want these invoices / usage statistics to be publicly exposed. 

There are some duplicates here: `GetBillingSummary` returns all the information in a single API call required for the Billing page. It is essentially combining `GetCurrentMonthUsage` `GetUnpaidBalance` `GetInvoiceHistory` - not sure if we want to make 3 async requests on the FE to get this data or just getting it all in 1 request is sufficient. I think the former may be easier to implement, but we can leave these proto defs in here to implement for down the line if we want to expose this.


proto/viam/app/v1/billing.proto
```diff
@@ -0,0 +1,103 @@
...
+  double cloud_storage_usage = 1;
+  double data_upload_usage = 2;
+  double data_egres_usage = 3;
+  double standard_compute_usage = 4;
+  double total_usage_with_discount = 5; // returns number with any discounts applied
+  double total_usage_without_discount = 6; // returns number without any discounts applied
+}
+
+message GetCurrentMonthUsageSummaryResponse {
+  double cloud_storage_usage = 1;
```
RoxyFarhad, line 31: essentially CurrentMonthUsageSummary - the former is used as part of the entire BillingSummary request that is retrieving info for the entire page.

proto/viam/app/v1/billing.proto
```diff
@@ -0,0 +1,103 @@
...
+  string org_id = 1;
+}
+
+message CurrentMonthUsageSummary {
+  double cloud_storage_usage = 1;
+  double data_upload_usage = 2;
+  double data_egres_usage = 3;
+  double standard_compute_usage = 4;
+  double total_usage_with_discount = 5; // returns number with any discounts applied
+  double total_usage_without_discount = 6; // returns number without any discounts applied
```
ehhong, line 27: should the discount be calculated separately? I feel like the name implies usage, billing independent

proto/viam/app/v1/billing.proto
```diff
@@ -0,0 +1,103 @@
...
+message GetBillingSummaryRequest {
+  string org_id = 1;
+}
+
+message GetBillingSummaryResponse {
+  CurrentMonthUsageSummary usage_summary = 1;
+  repeated InvoiceSummary invoices = 2;
+  double unpaid_balance = 3;
+  google.protobuf.Timestamp unpaid_balance_due_date = 4;
+  google.protobuf.Timestamp due_date = 5;
```
ehhong, line 101: what's the difference between `unpaid_balance_due_date` and `due_date`? is `due_date` for the most recent invoice and `unpaid_balance_due_date` is for the previous?

proto/viam/app/v1/billing.proto
RoxyFarhad: For the FE, we want to show the total usage with the discount and without. The actual numbers we show here are usage in terms of $ cost, not like GB usage etc. The name here can change but its referencing all the usage stats for the current month. The reason we shouldn't separate this out is because we calculate this on read i.e. these statistics are not computed every time there is an update. 

So if we wanted to separate this into with and without discounts, we would then be making 2 requests that would duplicate calls to the aggregator functions. One that applies a discount after and one that does not. This will strain the API for no reason, so doesn't make sense to me to separate it out.

proto/viam/app/v1/billing.proto
ehhong: what's the reason for not just using `CurrentMonthUsageSummary` here?

proto/viam/app/v1/billing.proto
RoxyFarhad: yup correct! This needs to be a bit more scoped because we don't currently turn off services if the balance isn't paid but im matching what is needed on design with this.

proto/viam/app/v1/billing.proto
ehhong: ah okay, I didn't realize that the usage was in terms of $ cost. will there be separate protos for usage in terms of GB usage/etc? if so, can we differentiate naming?

RoxyFarhad: Currently not needed in the designs, but i can change this to make it more clear its $ cost in case we want to add GB usage late

proto/viam/app/v1/billing.proto
ehhong: could we do `due_date` and `last_due_date`? was a bit confusing to figure out what the difference was here. also, does `unpaid_balance` include all charges up to the current moment that haven't been paid, or up to the most recent full month?

not super confident in this naming but maybe we can take inspiration from credit card statement balances?

proto/viam/app/v1/billing.proto
RoxyFarhad: The reasoning for the duplicate here is because we probably don''t need all the sub-requests here because we in theory want to get all the data for the single page `GetBillingSummaryRequest`, but i just left it in here in case we want to make requests tied to billing components rather than the page. 

I agree for API code readability sake it better to not have duplicate message structures here, but for the client it's better to not have nested layering, especially if the top layer has a single key.

proto/viam/app/v1/billing.proto
anjinai: Agree with Emily re: modeling this after credit card statement balances.

RoxyFarhad: yeah `due_date` and `oustanding_due_date` i think that last_due_date also sounds a bit unclear?

ehhong: what if we flipped it to be `current_due_date` and `invoice_due_date`?

RoxyFarhad: okay so based on credit card statements lets do:
statement balance - all unpaid balances at the end of the last billing cycle. 
current balance - statement_balance + current month usage so far. 

Speaking to Natalia about dates

proto/viam/app/v1/billing.proto
RoxyFarhad: updated!

proto/viam/app/v1/billing.proto
RoxyFarhad: updated except for the date -- for now, i added statement_balance_due_date + current_balance_due_date - but can change this down the line if this doesn't make sense?

ehhong: `statement_due_date` and `current_due_date`?

RoxyFarhad: `statement_balance_due_date` - when the pay the outstanding bill from last month if that is set
`current_balance_due_date` - when to pay the current bill (this should be the 15th of the Month)

The logic to actually writing these is still unclear, but im adding it in for now because its on the designs and we will want to show it.

RoxyFarhad: Okay so spoke to Natalia and we are going to remove the statement_balance_due_date. This should just be immediate i.e. the user should be prompted to just be paid immediately. 

The current_balance_due_date i.e. the day that the current balance is due is the 1st of the month i.e. when we charge if payment is set or the 15th of the current month. We may not need to pass this in from the BE, but leaving it in here for now.

proto/viam/app/v1/billing.proto
```diff
@@ -0,0 +1,114 @@
...
+  string org_id = 1;
+}
+
+message GetBillingSummaryResponse {
+  CurrentMonthUsageSummary usage_summary = 1;
+  repeated InvoiceSummary invoices = 2;
+  double statement_balance = 3; // all unpaid balances at the end of the last billing cycle
+  double current_balance = 4; // statement_balance + current_month_usage
+  double current_month_balance = 5; // current_month_usage_cost
+  google.protobuf.Timestamp current_balance_due_date = 7;
```
ehhong, line 112: being nitpicky here, but this does due date only refers to the current month's costs? technically everything before the current month balance is overdue. if so, can we rename to be `next_due_date`, `upcoming_due_date`, `current_month_due_date`, or something?

proto/viam/app/v1/billing.proto
```diff
@@ -0,0 +1,114 @@
...
+  google.protobuf.Timestamp due_date = 5;
+  repeated BillableResourceEvent items = 6;
+  string emailed_to = 7;
+}
+
+message GetCurrentMonthUsageSummaryRequest {
+  string org_id = 1;
+}
+
+message GetCurrentMonthUsageSummaryResponse {
```
ehhong, line 63: still feel like this would be worth making a single `CurrentMonthUsageSummary`, seems like you've done something similar below with `GetItemizedInvoiceResponse`? but your call

proto/viam/app/v1/billing.proto
RoxyFarhad: yup agreed! Will add

RoxyFarhad: updated!

ehhong (approved):

PR #239, opened 2023-04-05
RSDK-2476: Encoder Function in viamrobotics/api

martha-johnston:
Adds necessary encoder functions/files to fully add encoder to the api


penguinland (commented):
I've got two types of feedback: nitpicks about comments, and doubts about whether the scope doc is right in the first place. That second one is important, though perhaps I'm too late to change it and I need to move forward with what we already decided.

martha-johnston:
@penguinland I fixed all the formatting comments and will wait to hear what Fahmina has to say about the rest

penguinland (approved):
LGTM! I'm glad we discussed yesterday.

PR #244, opened 2023-04-12
RSDK-2385 expose human-readable device name and a unique identifier

bazile-clyde:
### Note
My changes are the few line in `camera.proto`. All other changes were auto-generated by running `make buf`.

### Description
This PR

- exposes a human-readable name for each webcam
- exposes an ID for each webcam just in case two webcams have the same human-readable name
- documents that the `Label` should now only be used internally for identifying cameras.

Here are two examples of some of the information this PR will expose to the front-end:
```
Name:  Laptop Camera: Laptop Camera
ID:  usb-0000:00:14.0-7
Label:  usb-Generic_Laptop_Camera_200901010001-video-index0
...
Name:  HD Web Camera: HD Web Camera
ID:  usb-0000:00:14.0-3
Label:  usb-HD_Web_Camera_HD_Web_Camera_Ucamera001-video-index0
```
As of now, the app displays the `Label` data to users. Both camera are attached to my Linux laptop. For Mac the ID will be a devices UID. Note there is not a typo or error in the name. This is exactly how those drivers chose to expose the names for their cameras. 

```
➜  api git:(RSDK-2385) v4l2-ctl --list-devices --all
HD Web Camera: HD Web Camera (usb-0000:00:14.0-3):
        /dev/video1
        /dev/video2
        /dev/media0

Laptop Camera: Laptop Camera (usb-0000:00:14.0-7):
        /dev/video3
        /dev/video4
        /dev/media1
```
### Testing
I tested this manually by making the necessary changes in `camera.proto`, running `make buf`, and running the RDK with my local version of this repo with these PR changes included. To make sure we would be sending the right info to the front end I printed out the information to my terminal from the discovery service function. The above info snippets are taking from that. To ensure there are no breaking changes with how we connect cameras, I made sure we can still see video streams in the app's control tab.


bhaney (approved):
LGTM

proto/viam/component/camera/v1/camera.proto
```diff
@@ -104,12 +104,16 @@ message Webcams {
...
+  // Camera driver label (for internal use only)
   string label = 1;
   // Camera driver status
   string status = 2;
   // Camera properties
   repeated Property properties = 3;
+  // Camera human-readable driver name
+  string name = 4;
+  // Camera unique identifier
+  string id = 5;
```
edaniels, line 116: maybe just unique_id. im assuming this value is permanent to the hardware and not just a bus? if it's not permanent, mabne we should comment that

edaniels (commented):
just 1 q but then lgtm

agreenb (approved):

proto/viam/component/camera/v1/camera.proto
bazile-clyde: It depends. As is, on Linux this will be the bus info and on Mac this is unique to each device. So it may be permanent or may not be. We could change this in the future too but as of now I'm just following the Linux Kernel docs for Linux and Martha followed AVFoundation's recommendations. I don't think it really matters what we use since it's only for disambiguating identical camera names. We'll still use the label for everything else.

edaniels: sound sgood!

edaniels (approved):

PR #245, opened 2023-04-17
change vision service API

bhaney:
This is a breaking change that uses the new vision service API as outlined in the ML Model Service Scope


proto/viam/service/vision/v1/vision.proto
bhaney, line 1: This was the only files that were touched, all others are generated from the make buf command

cheukt (approved):

PR #231, opened 2023-03-30
[RSDK-2241] remove temp map representation endpoints from SLAM

JohnN193:
This PR removes the temp map representation endpoints from the SLAM proto. These include `GetPointCloudMapStream`, `GetInternalStateStream`, `GetPositionNew` as well as associated messages. 

This PR can only be merged after these endpoints are removed from rdk and carto/orb and app is updated 
https://viam.atlassian.net/browse/RSDK-2246
https://viam.atlassian.net/browse/RSDK-2240
https://viam.atlassian.net/browse/RSDK-2256

ticket: https://viam.atlassian.net/browse/RSDK-2241

PR #247, opened 2023-04-20
[RSDK-2605] Document expectations of color field in PCDs

dmhilly:
This PR adds a short comment documenting how Viam interprets the color field in a PCD.

### DIFF
--- a/component/encoder/v1/encoder.pb.gw.go
+++ b/component/encoder/v1/encoder.pb.gw.go
@@ -15,7 +15,7 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/utilities"
-	"go.viam.com/api/common/v1"
+	v1_0 "go.viam.com/api/common/v1"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/grpclog"
@@ -247,7 +247,7 @@ var (
 )
 
 func request_EncoderService_DoCommand_0(ctx context.Context, marshaler runtime.Marshaler, client EncoderServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
-	var protoReq v1.DoCommandRequest
+	var protoReq v1_0.DoCommandRequest
 	var metadata runtime.ServerMetadata
 
 	var (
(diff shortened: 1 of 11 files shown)


proto/viam/service/slam/v1/slam.proto
```diff
@@ -67,7 +67,16 @@ message GetPointCloudMapResponse {
...
+  //
+  // Read more about the pointcloud format here:
   // https://pointclouds.org/documentation/tutorials/pcd_file_format.html
+  //
+  // Viam expects pointcloud data with fields "x y z" or "x y z rgb", and for
+  // this to be specified in the pointcloud header in the FIELDS entry. If color
+  // data is included in the pointcloud, Viam's services assume that the color
+  // value encodes a confidence score for that data point. Viam expects the
+  // confidence score to be encoded in the blue parameter of the RGB value, on a
+  // scale from 1-100.
```
JohnN193, line 79: Only question I have is should this description also talk about coordinate system assumptions for maps, but I think if we want to add that here then it should be a part of the documentation changes @jeremyrhyde will make for the coordinate change scope

JohnN193 (approved):
LGTM. Only thought is if coordinate assumptions should be documented here, but I think its fine to leave it out for this PR

proto/viam/service/slam/v1/slam.proto
dmhilly: Good point, I'd need some more context on our coordinate system assumptions, but from what I understand that's going to change with Jeremy's work, right? If so, I think we should wait until that's resolved before documenting.

EshaMaharishi: Agree, I think it's okay to just document this for now.

component/encoder/v1/encoder.pb.gw.go
```diff
@@ -15,7 +15,7 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/utilities"
-	"go.viam.com/api/common/v1"
```
EshaMaharishi, line 18: @JohnN193 or @jeremyrhyde I know we discussed why these unrelated files get changed when we change slam.proto in https://github.com/viamrobotics/api/pull/231/files#diff-2170dff6c9599d02586a135ae17ca0a4d9e2df172c7ab24ef4a04829aad8a101. Do you remember why and mind reminding me?

EshaMaharishi (approved):
LGTM assuming it's correct for the .pb.gw.go files to change!

nicksanford (approved):

PR #249, opened 2023-04-24
RSDK-2806: Movement sensor distance values should be in meters

stevebriskin:


cheukt (approved):

proto/viam/component/movementsensor/v1/movementsensor.proto
```diff
@@ -149,7 +149,7 @@ message GetAccuracyRequest {
 }
 
 message GetAccuracyResponse {
-  map<string, float> accuracy_mm = 1;
+  map<string, float> accuracy = 1;
```
penguinland, line 152: Discussion question: the old name included the units. Should the new name include units, too? Like `accuracy_m` so it's clear that it's not percentages.

penguinland (approved):
LGTM. 

What's the plan for communicating this to the rest of the world? Changing the field names within protobufs is a pretty major breaking change, and I imagine people who are already using the accelerometers on their rovers will need to change their code to match.

penguinland:
(Just to be clear: I reviewed the .proto file but skipped all the .ts and .go ones. If you have manually changed anything in the other files, I won't have noticed.)

proto/viam/component/movementsensor/v1/movementsensor.proto
stevebriskin: having unit in accuracy was a mistake. accuracy may refer to a variety of units. for example, current usage is limited to gps which are unitless values as far as i can tell.

penguinland: I had thought the GPS readings were accurate to within a couple meters (or a couple centimeters, if you have the RTK correction data), and these units made sense (the accuracy is measured in millimeters). You make a good point, though, that the gyroscope accuracy is likely in degrees per second, so the name of the map itself cannot describe the units of its values. Instead, the keys in the map could describe the units in their corresponding value. I do think it's important that units are associated with the data in some way.

penguinland: (but the keys are in the RDK repo, so no further changes are needed in this PR)

PR #250, opened 2023-04-24
convert bbox coordinates to double to match mongo

sbal13:
Bumping floats -> doubles!

Corresponding change in `app` https://github.com/viamrobotics/app/pull/1518/files


tahiyasalam (commented):
~~This change makes logical sense to me, but I'm hoping to get a bit more context on why we need to this, either in the PR or off-PR! Does this mean we would also need to updated the `DBAnnotation` struct in app?~~ Edit: see [thread](https://viaminc.slack.com/archives/C03JULLTDC1/p1682370524744419) for context

tahiyasalam (approved):
LGTM pending corresponding app changes!

agreenb (approved):
LGTM. For context, we made the scope changes [here](https://docs.google.com/document/d/1ZBOWZo0Y2RdfB07h2sMwN1KlElSQ_3ybMDg2y65VU2s/edit#) to make this consistent with the rest of the Viam API (we represent decimal values as double 99 other times vs only 3 as float), and to ensure that we can pass along the values to be stored as float64 in MongoDB (we were previously storing as float32 and would get a truncate error when making an API call from JS).

Could you attach the corresponding app change into this PR description? If possible, would be great to split that app change out into just the changes required here. Thank you!

agreenb (approved):
Re-approving to re-trigger the proto compilation

agreenb:
@edaniels Adding you if it's possible to force-push these changes; details [here](https://github.com/viamrobotics/api/pull/250#pullrequestreview-1398837721) and in email titled 'Bounding Box Labeling Scope Update'

PR #253, opened 2023-04-27
APP-1690 Billing API cleanup, add payment method

ehhong:
- removing unused endpoints
- add payment method to `GetBillingSummaryResponse`
    - only supporting cards for now


RoxyFarhad (approved):
LGTM - just revert rpcs not to break

PR #256, opened 2023-05-04
[RSDK-2925] Add MoveOnGlobe and supporting messages to API

spieswl:


biotinker:
I think you need to add the generated go files as well, no?

spieswl:
Automatically handled by an action now!

proto/viam/service/motion/v1/motion.proto
```diff
@@ -81,6 +87,31 @@ message MoveOnMapResponse {
...
+  // Optional linear velocity to target when moving
+  optional float linear_meters_per_sec = 7;
+  // Optional angular velocity to target when turning
+  optional float angular_deg_per_sec = 8;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
+}
+
+message MoveOnGlobeResponse {
+  bool success = 1;
```
Otterverse, line 112: Is there a case where there's a failure but not an error? If not, this probably isn't needed. Though I see this pattern is repeated throughout this file already, so I'm not gonna object to this instance.

Otterverse (approved):
LGTM

proto/viam/service/motion/v1/motion.proto
spieswl: Yeah, we may have preempted or stopped the Motion, for one example.

PR #257, opened 2023-05-09
APP-1470 Add rename robot part endpoint

ehhong:
Creating a lightweight endpoint to support renaming without needing to pass an entire robot part. This will also help fix a bug where `fqdn` is not updated on rename in the code sample.

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -95,6 +95,9 @@ service AppService {
   // Update a robot
   rpc UpdateRobotPart(UpdateRobotPartRequest) returns (UpdateRobotPartResponse);
 
+  // Rename a robot part
+  rpc RenameRobotPart(RenameRobotPartRequest) returns (RenameRobotPartResponse);
+
   // Create a new robot part
   rpc NewRobotPart(NewRobotPartRequest) returns (NewRobotPartResponse);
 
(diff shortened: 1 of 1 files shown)


michaellee1019:
Lets discuss this a bit more before adding. APIs are externally facing and hard to eliminate once introduced. Are we sure there isn't a way to handle the frontend better without introducing a new API:

1. Use events to trigger a rerendering of the code samples section.
2. Try using $computed?
3. Brute force would be to refresh the page on part name update.

ehhong:
> Lets discuss this a bit more before adding. APIs are externally facing and hard to eliminate once introduced. Are we sure there isn't a way to handle the frontend better without introducing a new API:
> 
> 1. Use events to trigger a rerendering of the code samples section.
> 2. Try using $computed?
> 3. Brute force would be to refresh the page on part name update.

sorry didn't update this in time. had some discussions with @RoxyFarhad and @DTCurrie and I think I will just work with the existing `UpdateRobotPart` endpoint. it's a bit heavy duty but introducing more granular endpoints like this seems like a larger discussion about API design and architecture. I'm adding a ticket to rethink this in the [API performance improvements](https://viam.atlassian.net/browse/APP-1299?atlOrigin=eyJpIjoiOTRjNWQ3NDc4MmJhNGI3N2E1YjY3MzgxNzI5NDhkYjgiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) epic.

PR #260, opened 2023-05-15
APP-1698 - Breakout billing page requests into multiple rather than single page. 

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-1698

This adds additional message types to support breaking `GetBillingSummary` into multiple smaller chunks.

Important notes:
- While adding `GetPaymentMethodCardRequest`, I realized that we may want a more general `GetPaymentMethodRequest` that contains the type of transaction (denoted `suggestion 2` in the initial commit). I am personally in favor of `suggestion 1` for this PR but I thought I should include the second suggestion in case we wish to minimize proto changes.


proto/viam/app/v1/billing.proto
```diff
@@ -105,10 +106,61 @@ message GetItemizedInvoiceResponse {
   Invoice invoice = 1;
 }
 
+// Begin Suggestion 1:
+message GetPaymentMethodCardRequest {
+  string org_id = 1;
+}
+
+message GetPaymentMethodCardResponse {
```
RoxyFarhad, line 114: We probably don't need GetPaymentMethodResponse and GetPaymentMethodCardResponse as separate rpc calls. The GetPaymentMethodResponse with a type parameter and maybe an optional object representing a card inside of that as you have done below looks good!

That being said, we may want something a bit more verbose than GetPaymentMethod as an rpc call, but something that gets all the stripe <> org information. Apart from the payment details, the rest of the stripe objects are defined on the org model but I think we will want to return billing_email in the same request here. 

A nice way of thinking about it is where do the rpc calls hit, so if we require looking at org table to get payment info (we need to get the stripe customer id to then get the payment info from stripe), pulling the billing email there also makes sense.

proto/viam/app/v1/billing.proto
```diff
@@ -105,10 +106,61 @@ message GetItemizedInvoiceResponse {
...
+  // defined if type is PAYMENT_METHOD_TYPE_CARD
+  optional PaymentMethodCard method = 2;
+}
+// End Suggestion 2
+
+
+message GetBalanceSummaryRequest {
+  string org_id = 1;
+}
+message GetBalanceSummaryResponse {
```
RoxyFarhad, line 143: The original balance field terms are also slightly outdated, and because they don't map exactly to the FE they are slightly confusing i would probably change it to something like:
statement_balance -> oustanding balance 
current_balance -> current_month_balance 

The usage breakdowns also would be good here.

proto/viam/app/v1/billing.proto
```diff
@@ -105,10 +106,61 @@ message GetItemizedInvoiceResponse {
...
+  // all unpaid balances at the end of the last billing cycle
+  double statement_balance = 1;
+  // statement_balance + current_month_balance
+  double current_balance = 2;
+  // equivalent to GetCurrentMonthUsageSummaryResponse.total_usage_with_discount
+  double current_month_balance = 3;
+  // due date for current month usage
+  google.protobuf.Timestamp current_month_due_date = 4;
+  string invoice_email = 5;
```
RoxyFarhad, line 152: would move this with the payment method request.

proto/viam/app/v1/billing.proto
RoxyFarhad: We also need the total_cost_with_discount, total_cost_without_discount, and discount here too.

proto/viam/app/v1/billing.proto
```diff
@@ -14,6 +14,7 @@ service BillingService {
   rpc GetBillingSummary(GetBillingSummaryRequest) returns (GetBillingSummaryResponse);
 }
 
+// Will be deprecated/removed in favor of GetCurrentMonthUsageSummaryResponse
```
RoxyFarhad, line 17: For the tickets also mark deprecated, could you please create and link a ticket,

RoxyFarhad (changes requested):

proto/viam/app/v1/billing.proto
```diff
@@ -80,35 +94,83 @@ message GetCurrentMonthUsageSummaryResponse {
...
+
+message GetCurrentMonthUsageResponse {
+  google.protobuf.Timestamp start_date = 1;
+  google.protobuf.Timestamp end_date = 2;
+  double cloud_storage_usage_cost = 3;
+  double data_upload_usage_cost = 4;
+  double data_egres_usage_cost = 5;
+  double remote_control_usage_cost = 6;
+  double standard_compute_usage_cost = 7;
+  double discount_amount = 8;
```
zaporter-work, line 109: new

proto/viam/app/v1/billing.proto
```diff
@@ -80,35 +94,83 @@ message GetCurrentMonthUsageSummaryResponse {
   double total_usage_without_discount = 11; // returns amt without any discounts applied
 }
 
+message GetCurrentMonthUsageRequest {
+  string org_id = 1;
+}
+
+message GetCurrentMonthUsageResponse {
+  google.protobuf.Timestamp start_date = 1;
+  google.protobuf.Timestamp end_date = 2;
```
zaporter-work, line 103: new

proto/viam/app/v1/billing.proto
zaporter-work: I have restructured this to be include previous invoices and to focus primarily on the top-banner and the invoice list. I have kept that month-based info in `GetCurrentMonthUsage`. How do you feel about this approach?

proto/viam/app/v1/billing.proto
```diff
@@ -80,35 +94,83 @@ message GetCurrentMonthUsageSummaryResponse {
...
+  double cloud_storage_usage_cost = 3;
+  double data_upload_usage_cost = 4;
+  double data_egres_usage_cost = 5;
+  double remote_control_usage_cost = 6;
```
RoxyFarhad, line 107: We aren't actually showing this now but I am happy to leave this in as we will show it eventually and then its hardcoded in the BE and not in the FE. Just remember in your next PR to hardcode this to 0.0 for now.

proto/viam/app/v1/billing.proto
```diff
@@ -80,35 +94,83 @@ message GetCurrentMonthUsageSummaryResponse {
...
+// TODO(APP-1865) may want to remove
 message GetItemizedInvoiceResponse {
   Invoice invoice = 1;
 }
 
+message GetBillingInformationRequest {
+  string org_id = 1;
+}
+
+message GetBillingInformationResponse {
```
RoxyFarhad, line 149: nit: maybe OrgBillingInfo? This entire page is BillingInfo really so having Org as a prefix makes it clear its more admin-related details.

RoxyFarhad (commented):
Looks good - almost there! Some small nits re name changing but otherwise i think its a clear breakdown

RoxyFarhad (commented):
Looks good - almost there! Some small nits re name changing but otherwise i think its a clear breakdown

proto/viam/app/v1/billing.proto
zaporter-work: +1. Yup, I added it because I saw it on the figma and thought it would be useful to have going forward.

proto/viam/app/v1/billing.proto
```diff
@@ -12,8 +12,15 @@ service BillingService {
   rpc GetInvoiceHistory(GetInvoiceHistoryRequest) returns (GetInvoiceHistoryResponse);
   rpc GetItemizedInvoice(GetItemizedInvoiceRequest) returns (GetItemizedInvoiceResponse);
   rpc GetBillingSummary(GetBillingSummaryRequest) returns (GetBillingSummaryResponse);
+  // Detailed breakdown of current month's costs
+  rpc GetCurrentMonthUsage(GetCurrentMonthUsageRequest) returns (GetCurrentMonthUsageResponse);
+  // Org-level information (like billing email and payment details)
+  rpc GetBillingInformation(GetBillingInformationRequest) returns (GetBillingInformationResponse);
+  // Total outstanding balance and previous invoices
+  rpc GetBalanceSummary(GetBalanceSummaryRequest) returns (GetBalanceSummaryResponse);
```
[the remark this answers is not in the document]
zaporter-work: Amusingly, I originally had it as `GetInvoiceSummary`, however, that was confusing because it meant that `GetInvoiceSummary` returned a list of `InvoiceSummary` objects. I am happy to put it back as `GetInvoiceSummary`, however there is some naming confusion there.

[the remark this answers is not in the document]
RoxyFarhad: GetInvoicesSummary ? making it plural makes it clear a list of them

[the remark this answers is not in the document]
zaporter-work: I am happy with that. Changed.

proto/viam/app/v1/billing.proto
zaporter-work: Agreed. Changed.

RoxyFarhad (approved):
LGTM! Congrats on your first fleet API ticket :)

PR #261, opened 2023-05-16
Add frame_rate to camera property

SeanAvery:
This PR simply adds `frame_rate` to Property message in camera.proto for use in App webcam builder.


bhaney (approved):
LGTM, make sure to call `make lint` and `make dist/buf` before merging

PR #262, opened 2023-05-19
APP-1561: Add getInvoicePdf proto

mcous:
WIP branch for public API changes in support of APP-1561


ehhong (approved):
lgtm! out of curiosity, does the response automatically download the PDF? or does the frontend have to process the `GetInvoicePdfResponse` to then download it?

RoxyFarhad (approved):
lgtm!

mcous:
> lgtm! out of curiosity, does the response automatically download the PDF? or does the frontend have to process the GetInvoicePdfResponse to then download it?

I haven't figured out how to get the PDF to donwload automatically, yet. There might be a header setting that makes this happen, so I'll continue playing with it.

For now, though, the streaming response makes it pretty easy for the front-end to construct a `Blob` that the client can then download through a data URL

1. Collect streaming file response:
https://github.com/viamrobotics/app/blob/a476253396d4c949549ad19c992f94653747ffb4/frontend/src/lib/api/billing.ts#L219-L238
2. Download:
https://github.com/viamrobotics/app/blob/a476253396d4c949549ad19c992f94653747ffb4/frontend/src/pages/orgs/billing.vue#L99-L107

PR #263, opened 2023-05-22
Add contributing docs and clean up Makefile

mcous:
## Overview

While working on #262, I noticed a few little issues with the `Makefile`. I see these same issues across the codebase, so I think it's a case of the same thing getting copy-pasted across multiple projects.

This PR resolves those issues and starts a `Contributing` section in the README because I was confused about what to do the first time I dropped into this library.

## Review requests

Do these changes make sense?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,26 @@
-TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
+TOOL_BIN := bin/gotools/$(shell uname -s)-$(shell uname -m)
+PROTO_FILES := $(shell find proto/ -type f -name '*.proto')
 
-PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):`pwd`/node_modules/.bin:${PATH}"
+export GOBIN := $(shell pwd)/$(TOOL_BIN)
+export PATH := $(GOBIN):$(shell npm root)/.bin:$(PATH)
 
-PROTO_FILES=$(shell find proto/ -type f -name '*.proto')
+.PHONY: all
+all: clean build
 
+.PHONY: setup
 setup:
 	bash etc/setup.sh
 
-clean-all:
+.PHONY: build
+build: dist/buf-go dist/buf-web
+
+.PHONY: clean
+clean:
 	git clean -fxd
 
 dist/tool-install: Makefile
 	npm ci --audit=false
-	GOBIN=`pwd`/$(TOOL_BIN) go install google.golang.org/protobuf/cmd/protoc-gen-go \
+	go install google.golang.org/protobuf/cmd/protoc-gen-go \
 		github.com/bufbuild/buf/cmd/protoc-gen-buf-breaking \
 		github.com/bufbuild/buf/cmd/protoc-gen-buf-lint \
 		github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,18 +1,26 @@
-TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
+TOOL_BIN := bin/gotools/$(shell uname -s)-$(shell uname -m)
+PROTO_FILES := $(shell find proto/ -type f -name '*.proto')
 
-PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):`pwd`/node_modules/.bin:${PATH}"
```
mcous, line 3: Makefiles do not support backticks. This works sort-of by accident: the backticks are inserted verbatim into `bash`, which then expands them.

This is brittle, so I replaced `pwd` with `$(shell pwd)`, which Makefile will properly expand itself. I also switch the `=` (evaluate every use) with `=:` (evaluate at assignment)

Makefile
```diff
@@ -1,18 +1,26 @@
...
+export GOBIN := $(shell pwd)/$(TOOL_BIN)
+export PATH := $(GOBIN):$(shell npm root)/.bin:$(PATH)
 
-PROTO_FILES=$(shell find proto/ -type f -name '*.proto')
+.PHONY: all
+all: clean build
```
mcous, line 8: By convention, I think running `make` without arguments should run an `all` target, and a simple `clean build` seemed like the way to go with that

Makefile
```diff
@@ -1,18 +1,26 @@
-TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
+TOOL_BIN := bin/gotools/$(shell uname -s)-$(shell uname -m)
+PROTO_FILES := $(shell find proto/ -type f -name '*.proto')
 
-PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):`pwd`/node_modules/.bin:${PATH}"
+export GOBIN := $(shell pwd)/$(TOOL_BIN)
+export PATH := $(GOBIN):$(shell npm root)/.bin:$(PATH)
 
-PROTO_FILES=$(shell find proto/ -type f -name '*.proto')
+.PHONY: all
```
mcous, line 7: We were missing `PHONY` on all our phony targets

edaniels:
@Otterverse  is better at Makefile's than I am so I'm swapping to him :) Sorry James!

stevebriskin (approved):
nice! learned some new things reviewing this. we should probably do this for Makefiles in other repos too

mcous:
Go modules changed and I would not expect them to. Setting PR to a draft while I investigate

Makefile
```diff
@@ -26,23 +31,26 @@ dist/tool-install: Makefile
...
+	touch dist/buf-web
 
+.PHONY: lint
 lint: dist/tool-install
-	PATH=$(PATH_WITH_TOOLS) buf lint
-	PATH=$(PATH_WITH_TOOLS) buf format -w
+	buf lint
+	buf format -w
 	export pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$$pkgs" | xargs go vet -vettool=$(TOOL_BIN)/combined
 	export GOGC=50 pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$$pkgs" | xargs $(TOOL_BIN)/golangci-lint run $(LINT_BUILD_TAGS) -v --fix --config=./etc/.golangci.yaml
```
Otterverse, line 56: If you remove TOOL_BIN you can remove it here too. just "combined" and "golangci-lint" should work since it'll be in the PATH now.

Otterverse (changes requested):
Submitting a review anyway, even though you went back to draft. I'll DM you on slack about the issues with the proto generation happening still.

cheukt:
generally looks good to me, removing myself from reviewers since I'm not much good at Makefiles either

Makefile
mcous: As discussed in Slack, leaving these as per-target `.PHONY`'s as recommended by [this Makefile styleguide](https://clarkgrubb.com/makefile-style-guide) to make individual phony targets easier to grok/modify/remove

Makefile
```diff
@@ -26,23 +30,26 @@ dist/tool-install: Makefile
...
 
+.PHONY: lint
 lint: dist/tool-install
-	PATH=$(PATH_WITH_TOOLS) buf lint
-	PATH=$(PATH_WITH_TOOLS) buf format -w
-	export pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$$pkgs" | xargs go vet -vettool=$(TOOL_BIN)/combined
-	export GOGC=50 pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$$pkgs" | xargs $(TOOL_BIN)/golangci-lint run $(LINT_BUILD_TAGS) -v --fix --config=./etc/.golangci.yaml
+	buf lint
+	buf format -w
+	export pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$$pkgs" | xargs go vet -vettool=`which combined`
```
mcous, line 54: `-vettool=combined` produced the following error

```shell
export pkgs=`go list -f '{{.Dir}}' ./... | grep -v gen | grep -v proto` && echo "$pkgs" | xargs go vet -vettool=combined
go: can't execute /Users/mc/viam/api/combined -flags: fork/exec /Users/mc/viam/api/combined: no such file or directory
```

I decided to go with `` `which combined` `` to enable the requested removal of the `TOOL_BIN` variable, going off an example from the `go vet` docs. Also happy to bring `TOOL_BIN` back into the mix, if that's preferable

```shell
$ go help vet
...
The -vettool=prog flag selects a different analysis tool with alternative
or additional checks.
For example, the 'shadow' analyzer can be built and run using these commands:

  go install golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
  go vet -vettool=$(which shadow)
```

mcous (commented):
@Otterverse change requests addressed

Otterverse (approved):
LGTM!

PR #265, opened 2023-05-24
DATA-1352 - Add error to TrainingJobMetadata in API

tahiyasalam:


proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -62,6 +62,12 @@ enum TrainingStatus {
   TRAINING_STATUS_CANCELING = 6;
 }
 
+message ErrorStatus {
+  // A Code is an unsigned 32-bit error code as defined in the gRPC spec.
+  uint32 code = 1 [(tagger.v1.tags) = "bson:\"code\" json:\"code,omitempty\""];
```
AaronCasas, line 67: Should we use the code enum type directly here? https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

tahiyasalam: Originally, I tried to reference the status type directly, but ran into an error. I saw this [link](https://stackoverflow.com/questions/60878404/referencing-external-protos-like-google-rpc-status-proto) for referencing external protos. Do you have a suggestion for referencing either code or status?

AaronCasas: Hmm yeah that answer seems to indicate you need a local copy, but I do see we have existing protos that import things and I don't think provide a local copy (e.g. `google/protobuf/struct.proto"`, `"google/api/annotations.proto`). Does `import "google/rpc/code.proto";` work?

tahiyasalam: It works! I think originally I was importing an incorrect proto.

AaronCasas (approved):
🚀

PR #267, opened 2023-06-02
Specify shell to make sure macOS `make` respects `PATH`

mcous:
This PR adds `SHELL := bash` to the Makefile because without it macOS `make` does not respect `export PATH`. Anecdotally, this is not [the first time I've seen this](https://github.com/Opentrons/opentrons/pull/4809). For fun, also see [this Stack Overflow post from 2014](https://stackoverflow.com/questions/21708839/problems-setting-path-in-makefile/21709821#21709821)!

This is probably a good thing to do by default in our Makefiles anyway. For example, it's part of [this style guide's default prologue](https://clarkgrubb.com/makefile-style-guide#prologue). From that styleguide, setting `.SHELLOPTIONS` seems like an potentially good move to consider, too

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL := bash
+
 PROTO_FILES := $(shell find proto/ -type f -name '*.proto')
 
 export GOBIN := $(shell pwd)/bin/gotools/$(shell uname -s)-$(shell uname -m)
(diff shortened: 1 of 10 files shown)


Otterverse (approved):
LGTM

PR #273, opened 2023-06-12
[RSDK-3404] Add comment to camera proto regarding intrinsic and distortion parameters for transform cameras

vpandiarajan20:
This PR adds a short comment to the camera proto file because users seemed to be confused when transform cameras errored if they were passed intrinsic and distortion of the underlying camera. These underlying parameters are not used for transforms and they should be left empty. The transforms will then use an empty cameraModel for initialization and function properly.


bhaney:
another thing you'll want to do it make all the derived proto files, too

vpandiarajan20:
Yeah, I think I get it a little bit better now, I was under the assumption that users were only running into this for transform cameras, but the error can occur any time that the camera's parameters are set to zero, right?

PR #275, opened 2023-06-16
RSDK-3644: Gantry API changes

martha-johnston:
1. expose a way to set axes speeds: MoveToPosition with Velocity list
2. add home function to api


oliviamiller (approved):

proto/viam/component/gantry/v1/gantry.proto
```diff
@@ -26,6 +26,13 @@ service GantryService {
     };
   }
 
+  // Home runs the homing sequence of a gantry and returns true once it's completed.
+  rpc Home(HomeRequest) returns (HomeResponse) {
+    option (google.api.http) = {
```
randhid, line 31: I'm fine with this, but we might need to add a heartbeat to this as it is an actuating method. @maximpertsov advise when you see this. We'll address it after it is merged.

maximpertsov: it depends on what behavior we want. if we want the homing sequence to potentially stop before it completes whenever the client that requested this command is disconnected, then yes we need to make this operation "safety monitored" by adding the following line:

```suggestion
    option (common.v1.safety_heartbeat_monitored) = true;
    option (google.api.http) = {
```

however, maybe we want the homing sequence to complete even if the client gets disconnected - in that case we can leave as is.

PR #276, opened 2023-06-22
RSDK-3560 Add GetLatestMapInfo to SLAM Proto

PStrutz:
JIRA ticket: https://viam.atlassian.net/browse/RSDK-3560.


proto/viam/service/slam/v1/slam.proto
```diff
@@ -85,7 +93,7 @@ message GetPointCloudMapResponse {
 }
 
 message GetInternalStateRequest {
-  // Name of slam service
```
EshaMaharishi, line 88: Why make this plural?

proto/viam/service/slam/v1/slam.proto
```diff
@@ -94,3 +102,13 @@ message GetInternalStateResponse {
   // mapping/localization
   bytes internal_state_chunk = 1;
 }
+
+message GetLatestMapInfoRequest {
+  // Name of the SLAM algo
+  string name = 1;
```
EshaMaharishi, line 108: Will this be different from what's returned in GetInternalState?

EshaMaharishi (commented):
Left a couple questions, but in general feel free to just have Jeremy review PR's for this project!

proto/viam/service/slam/v1/slam.proto
PStrutz: Yep! From what I understand, GetInternalState retrieves all the information needed to reconstruct the point cloud (e.g. how many different chunks make up the PC, etc.) - although I am not 100% certain what EXACTLY this function returns, as this was already in the code before I worked on it. GetLatestMapInfoRequest is the new function I added to track the timestamp of the last map that was created.

EshaMaharishi: Ah, I misread this and thought it was being returned in the response. It makes sense to have it in the request!

jeremyrhyde (approved):
LGTM. Cleanly implemented!

PR #282, opened 2023-06-28
APP-2112: Modular Registry Protos

michaellee1019:


proto/viam/app/v1/robot.proto
```diff
@@ -256,12 +256,14 @@ message RemoteAuth {
 
 message AgentInfo {
```
RoxyFarhad, line 257: nit: I saw that we are making AgentInfo from optional -> required in the ConfigRequest. Should we make a note here to do that once we have the setup to always send that info in the request?

RoxyFarhad (approved):
just small clarification q but otherwise lgtm! 

I saw that adding `type` here:

```
message PackageConfig {
  string name = 1;
  string package = 2;
  string version = 3;
  string type = 4;
}

```
was also in the tech spec. Is this something we need to add here too?

proto/viam/app/v1/robot.proto
michaellee1019: I should have commented about this my bad. I tried this and its more complicated IMO, I'll tackled it after some other parts of the project are done. I wanted to move quicker than think of deprecation so I created a separate ticket for this yesterday: https://viam.atlassian.net/browse/APP-2142

proto/viam/app/v1/app.proto
```diff
@@ -736,3 +748,122 @@ message AuthorizedPermissions {
...
+  // The id of the module, containing the namespace and name
+  string module_id = 1;
+  // The name of the module
+  string name = 2;
+  // The visibility of the module
+  Visibility visibility = 3;
+  // The versions of the module that are available
+  repeated VersionHistory versions = 4;
+  // The url to reference for documentation, code, etc.
+  string url = 5;
```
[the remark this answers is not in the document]
michaellee1019: Yes we changed the name to url at the last minute (tuesday's doc review meeting)

PR #284, opened 2023-06-29
APP-2140 Update authorizations endpoints for org settings page

ehhong:
endpoint to support this part of the org settings page:

image


proto/viam/app/v1/app.proto
```diff
@@ -708,18 +723,27 @@ message RemoveRoleRequest {
...
-  string identity_id = 1;
-  repeated Authorization authorizations = 2;
+message ListAuthorizationsResponse {
+  repeated Authorizations authorizations = 1;
 }
 
-message ListAuthorizationsResponse {
+message ListAuthorizationsForUserRequest {
+  string user_id = 1;
+  string organization_id = 3;
```
ehhong, line 743: actually, should we omit the query `organization_id` and use the current org on the state? or should we make this explicit

proto/viam/app/v1/app.proto
```diff
@@ -121,6 +121,9 @@ service AppService {
   // Get a list of robots
   rpc ListRobots(ListRobotsRequest) returns (ListRobotsResponse);
 
+  // Get a list of robots by organization
+  rpc ListRobotsByOrg(ListRobotsByOrgRequest) returns (ListRobotsByOrgResponse);
```
acrantel, line 125: Will `ListAuthorizationsForUser` already include all the robots that have permissions set for that user? If so, then do we still need `ListRobotsByOrg`?

ehhong: `ListRobotsByOrg` would still be useful for the select resource modal (where we load all robots a user has read access on regardless of current authorizations), but it's not strictly necessary because we can call `ListRobots` on every location we load.

I'll let @piokasar make the call on this one

proto/viam/app/v1/app.proto
piokasar: let's make it explicit - according to Michael, currentOrg isn't always populated correctly

proto/viam/app/v1/app.proto
piokasar: im tempted to say no because of what Michael said about reducing the number of endpoints we expose, but it would definitely cut down on the number of calls we make. Is there anywhere in the future where we would have to have the robots without needing the locations that would result in us using ListRobotsByOrg again?

proto/viam/app/v1/app.proto
piokasar, line 743: should be 2

ehhong: oops, will fix!

proto/viam/app/v1/app.proto
```diff
@@ -153,9 +156,13 @@ service AppService {
   // Deletes an IdentityAuthorization
   rpc RemoveRole(RemoveRoleRequest) returns (RemoveRoleResponse);
 
-  // Shows organization, location, and robot level permissions that exist on the resource
+  // Returns all authorization roles for a given resource
   rpc ListAuthorizations(ListAuthorizationsRequest) returns (ListAuthorizationsResponse);
 
+  // Returns a user's authorization roles in the org
+  rpc ListAuthorizationsForUser(ListAuthorizationsForUserRequest) returns (ListAuthorizationsForUserResponse);
```
piokasar, line 163: one thing to note when writing this - you only want to return the resources in each identity authorization that contains a resource that the current app user also has access to. We don't want to return info about resources that other users have access to that the current user doesn't have access to / shouldn't even know exist.

proto/viam/app/v1/app.proto
ehhong: i don't think so unless we add more robot views on the org settings page, but i don't see an immediate reason for that. i'm good with removing it

proto/viam/app/v1/app.proto
```diff
@@ -708,21 +712,30 @@ message RemoveRoleRequest {
...
   repeated Authorizations authorizations = 1;
 }
 
+message ListAuthorizationsForUserRequest {
+  string user_id = 1;
+  string organization_id = 2;
+}
+
+message ListAuthorizationsForUserResponse {
+  repeated Authorization authorizations = 1;
```
ehhong, line 736: note that the return of this is a `repeated Authorization` instead of a `repeated Authorizations`. if i'm understanding correctly i think it's okay to omit the `identity_id` part of the `Authorizations` because these should all be for the same identity (the given user)

ehhong:
@michaellee1019 hello, @piokasar and I are suggesting a few structural changes to the API, we know they are pretty substantial so please let us know what considerations we might be overlooking.
1. We suggest modifying the `ListAuthorizations` endpoint to be cumulative and return all authorizations within an organization, with the option of filtering by resource(s). 
2. We suggest modifying the proto for `Authorization` to be entirely flat. This is to make data parsing on the frontend simpler and reduce the logical dependency of understanding how authorizations are stored in the db. At the cost of repeating some data, we can also avoid `Authorization` and `Authorizations` objects, instead always returning a single `Authorization` or multiple authorizations.

please let us know what you think!

proto/viam/app/v1/app.proto
```diff
@@ -680,47 +681,38 @@ message DeleteRobotPartSecretRequest {
...
+  string resource_type = 2;
+  string resource_id = 3;
 }
 
-message AddRoleResponse {
+message AddRoleRequest {
   Authorization authorization = 1;
 }
 
+message AddRoleResponse {}
```
ehhong, line 701: wasn't sure why `AddRoleResponse` needed to return the added authorization, so I removed it. lmk if this should be reinstated

proto/viam/app/v1/app.proto
```diff
@@ -680,47 +681,38 @@ message DeleteRobotPartSecretRequest {
...
-  string authorization_type = 1;
-  string authorization_id = 2;
-  string resource_type = 3;
-  string resource_id = 4;
+  string identity_id = 1; 
+  string authorization_type = 2;
+  string authorization_id = 3;
+  string resource_type = 4;
+  string resource_id = 5;
+  string org_id = 6;
```
michaellee1019, line 689: Suggest to keep the numbers of existing fields as is. Just add the new fields with additional numbers. The numbers don't need to be in order. It will cause less backwards incompatible changes.

proto/viam/app/v1/app.proto
michaellee1019: Fine with it. I think we did this because is pretty standard to return the just created data as a response in our APIs. Not sure if we actually use this on the frontend, but will need to remove the logic on the backend to return just an empty response.

michaellee1019 (commented):
Agreed that these are the right choice :)

proto/viam/app/v1/app.proto
ehhong: ah if it's standard, should we leave it in?

ehhong: actually, going to delete it because the other `createX` cases seem to do some transformation on the request data

michaellee1019 (approved):

michaellee1019 (approved):

ehhong:
@stevebriskin can you merge when you get the chance? this will unblock frontend permissions work

michaellee1019 (approved):
Approving to kick off a new protos build (fixes merge conflicts)

michaellee1019 (approved):
Forgot to remove label.

piokasar (approved):

PR #285, opened 2023-06-30
Deprecate os field later

michaellee1019:
I didn't realize that deprecating an API field will cause linting errors for go code that depends on that. So undoing this change until we are ready for it (in a few weeks when both App and RDK are using `platform` field instead). Otherwise people will be blocked on upgrading the API until then.


DTCurrie (approved):
✅

michaellee1019:
> ✅

Thanks for the quick review!

PR #292, opened 2023-07-06
Modular Registry Proto Updates

michaellee1019:
Three changes to the initial proto commit for modular registry:
1. Added the missing `entrypoint` field to UpdateModule
2. Added `organization_id` field to Module type. We debated this in our spec, to omit it unless needed. I've found that its easier to have this field, even though the name includes the public_namespace which infers the organizaiton. However having the organization_id directly, means less database lookups.
3. As discussed today, a url is returned in Create/Update/Upload APIs that the user can then load to see the published module.


zaporter-work (approved):
LGTM. 

Side thought: We may eventually want to consider a `GetModuleInfo` or something API that returns url + stats + owner + whatever else rather than just appending the url to the response.

michaellee1019:
> Side thought: We may eventually want to consider a `GetModuleInfo`

I think that we can just make GetModule that API. And have that return all info. I think its better in this case to return some things to eliminate the need for a Get after a Create. Most of our APIs return the full object on create/update, and could do that instead.

PR #293, opened 2023-07-06
APP-2173 Update fragments endpoints to take in org id

ehhong:
clarification q:
- do fragments always belong to an org?
- if so, would a user ever access a fragment in an org they don't belong to?

an alternate design is to make these endpoints use the eventual `currentOrg` field stored on the `user` object in mongo -- but this will take longer to implement


DTCurrie:
> clarification q:
> 
> * do fragments always belong to an org?

That is my understanding.

> * if so, would a user ever access a fragment in an org they don't belong to?

If they were public, but I think only Viam can make public fragments right now. 

@nataliajacobowitz are both of the above correct?

> an alternate design is to make these endpoints use the eventual `currentOrg` field stored on the `user` object in mongo -- but this will take longer to implement

We can ticket this out as a mini project.

michaellee1019 (approved):

DTCurrie (approved):
✅

PR #294, opened 2023-07-07
APP-2173 Revert "Update fragments endpoints to take in org id (#293)"

ehhong:
This reverts commit ae0d4c06464b508062bf88c366ae54402eec6f58.

Talked with @piokasar and we decided to make fragments requests consistent with other resource requests and have the frontend filter for fragments that shouldn't be displayed in the current org


ehhong:
@stevebriskin could you merge this when you get the chance? I'm reverting some changes I made yesterday that haven't been implemented yet.

PR #295, opened 2023-07-07
APP-2173 Add org ID parameter for createFragment, remove readonly flag

ehhong:
- Modifying `createFragment` endpoint to take in an owner org ID
- Removing readonly flag from fragments and implementing readonly with our current permissions system

Implementation of new API is here: https://github.com/viamrobotics/app/pull/1994


piokasar (approved):

ehhong:
@stevebriskin could you merge this PR? it does contain a breaking change but I have a PR ready to test with this change https://github.com/viamrobotics/app/pull/1994

this removes reliance on the current org cookie for fragments

proto/viam/app/v1/app.proto
```diff
@@ -586,13 +586,12 @@ message Fragment {
   bool public = 5 [(tagger.v1.tags) = "bson:\"public\" json:\"public\""];
   google.protobuf.Timestamp created_on = 6 [(tagger.v1.tags) = "bson:\"created_on\""];
   string organization_name = 7;
-  bool readonly = 8;
   // number of robot parts using this fragment
-  int32 robot_part_count = 9;
+  int32 robot_part_count = 8;
```
michaellee1019, line 590: Hey @ehhong changing field numbers is not good practice. Over the network this data gets transmitted like 9=value,10=value to save space, and is the primary purpose of the numbers.

Instead just delete readonly field and don't renumber the rest.

ehhong: okay! thanks for the heads up, didn't realize the numbers didn't have to be consecutive. will make that change now

PR #296, opened 2023-07-10
Modular Registry proto updates

michaellee1019:
This PR encompasses the following changes:
1. Namespace is no longer required to be declared until making a module public. `organization_id` added as an optional field on all Module APIs. OrgId can be omitted from requests once a `module_id` string contains "namespace:name", otherwise is required.
2. Namespace is removed from CreateModule API. If an org has a namespace it will be resolved server side and put into the module_id. Thanks @zaporter-work for the idea, which simplifies the interaction between API and CLI.
3. Adds the missing `entrypoint` field on Module type.
4. Adds `platform` as an optional field on CreatePackage API, making it easy to extend the implementation to UploadModuleFile API.


proto/viam/app/v1/app.proto
```diff
@@ -858,6 +864,8 @@ message Module {
   // A list of models that are available in the module
   repeated Model models = 7;
   // The total number of robots using this module
+  // The executable to run to start the module program
+  string entrypoint = 11;
   int64 total_robot_usage = 8;
```
zaporter-work, line 869: [nit] This is cutting a comment/code pair in half

zaporter-work (approved):
LGTM!

After more thought, we don't need a `LookupByPublicNamespace` req just yet.

proto/viam/app/v1/app.proto
michaellee1019: Good catch, updated.

zaporter-work (approved):

michaellee1019:
Hi @stevebriskin can you merge in this breaking change. We are switching from `public_namespace` to `organization_id` since `namespace` is not required until the module is public.

PR #297, opened 2023-07-11
RSDK-3967 - Add motionconfiguration and update moveonglobe request

nfranczak:


proto/viam/service/motion/v1/motion.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package viam.service.motion.v1;
+package viam.service.motion.v2;
```
nfranczak, line 3: since moveonglobe request has changed that is breaking, and we need to bump the version, right?

edaniels: I don't think we're adopting that convention yet although the intent is appreciated :). We just accept breaking changes for motion right now, right @raybjork and @Fahmina?

nfranczak: okii, ill change back

raybjork: Yeah we did not plan to do this

edaniels (approved):
LGTM

raybjork (approved):
With the tech spec slowly reaching a stable place I think its ok to merge this now.

proto/viam/service/motion/v1/motion.proto
```diff
@@ -81,6 +81,23 @@ message MoveOnMapResponse {
...
+  // The name of the vision service(s) that will be used to detect obstacles
+  repeated common.v1.ResourceName vision_services = 1;
+  // Sets the frequency to poll for the position of the robot
+  optional double position_polling_frequency = 2;
+  // Sets the frequency to poll the vision service(s) for new obstacles
+  optional double obstacle_polling_frequency = 3;
+  // Sets the distance in meters that a robot is allowed to deviate from the motion plan
+  optional double plan_deviation_m = 4;
+  // If this value is set, it will set the radius of an ellipse that a robot should search in when attempting to replan.
+  optional double replan_cost_factor = 5;
```
raybjork, line 94: This is no longer going into the API.  We updated the scope doc but I forgot to let you know about it

raybjork (changes requested):

proto/viam/service/motion/v1/motion.proto
nfranczak: will remove right now

raybjork (approved):

proto/viam/service/motion/v1/motion.proto
```diff
@@ -94,10 +109,8 @@ message MoveOnGlobeRequest {
   common.v1.ResourceName movement_sensor_name = 5;
   // Obstacles to be considered for motion planning
   repeated common.v1.GeoObstacle obstacles = 6;
-  // Optional linear velocity to target when moving
-  optional float linear_meters_per_sec = 7;
-  // Optional angular velocity to target when turning
-  optional float angular_deg_per_sec = 8;
```
tessavitabile, line 100: These field names and indexes should be reserved, like here: https://github.com/viamrobotics/api/pull/312/files

PR #298, opened 2023-07-11
DATA-1659 Add GCS region field to org and location API

kaywux:
App proto changes for GCS regions functionality. Corresponding app PR for breaking change to make UpdateOrganizationRequest name optional: https://github.com/viamrobotics/app/pull/2020

Ticket: https://viam.atlassian.net/browse/DATA-1659.


proto/viam/app/v1/app.proto
```diff
@@ -258,8 +261,10 @@ message GetOrganizationResponse {
 
 message UpdateOrganizationRequest {
   string organization_id = 1;
-  string name = 2;
+  optional string name = 2;
```
kaywux, line 264: It looks like the UpdateOrganizationRequest was previously only used to change an org's name; any objections to making this change now that we're using this update request for other fields? @michaellee1019

kaywux: If this change is okay I can go ahead and make the corresponding app changes as well.

michaellee1019 (approved):
Ok with it. I like the idea of our Update APIs having all optional fields, updating only what is necessary at the time. The design of the org setting page really warrants this change, given that each setting has its own box with an update button next to it (as opposed to a single form with all fields).

Just make sure you follow up with app changes quickly. There is a lot of velocity in app right now and this will block other API changes that happen afterwards.

agreenb (approved):

michaellee1019 (approved):
Approving again to rebuild protos without conflicts

kaywux:
@stevebriskin could you merge in this change? Breaking change is that the `name` field in `UpdateOrganizationRequest` is now optional; will follow this up with [app changes](https://github.com/viamrobotics/app/pull/2020).

PR #300, opened 2023-07-12
add UpdateInviteRequest rpc method, change names of OrganizationInvite method/messages to Invite

piokasar:
We need to let people update invites rather than send them one by one until a user has accepted. When we add an invite to the invite collection, we'll also store the current app user as the inviter and delete invites from the collection on accepting rather than sending.


michaellee1019:
> I'm not sure why that endpoint isn't being used now.

We probably didn't have the API at that point.

michaellee1019:
One last gut check before approving: If we are only having one API, it should still be named Update instead of AddAndRemove? Add/Remove makes sense if two API that you need to differentiate. We try to use Update unless some other action is more accurate. I don't feel strongly either way, except just know we need to stick with our decision long term.

piokasar:
> One last gut check before approving: If we are only having one API, it should still be named Update instead of AddAndRemove? Add/Remove makes sense if two API that you need to differentiate. We try to use Update unless some other action is more accurate. I don't feel strongly either way, except just know we need to stick with our decision long term.

That makes sense - I'll change it to update.

michaellee1019 (approved):

piokasar:
@michaellee1019 I think we might have to rebuild the protos somehow? I don't think solving this merge conflict manually would be great

ehhong (approved):

PR #301, opened 2023-07-13
Fix package type, add GetOrganizationNamespaceAvailability

michaellee1019:
Mistakenly used a string instead of PackageType enum in GetPackage. Adding a new GetOrganizationNamespaceAvailability API to check for namespace availability.


RoxyFarhad (approved):
think u will need steve/eric to force merge this

agreenb (approved):
LGTM for GetPackageRequest, thanks for the fix!

michaellee1019:
Hi @stevebriskin, can you merge this PR? I changed the type of `type` to be the correct enum instead of string. Even though the GetPackage API is used in the RDK, the field is not currently passed (will be soon) so there is not impact to this breaking change.

PR #307, opened 2023-07-25
RSDK-4055 Add PowerSensor component 

oliviamiller:
scope doc: https://docs.google.com/document/d/1HjShXMyD6HQkZkAo0bvDe61STdxxfyBrVnbRC6nSlM0/edit


martha-johnston (changes requested):
Code looks good, but I would add comments for each function just to be consistent with all the other component proto files

martha-johnston (approved):

PR #313, opened 2023-07-31
APP-2311 Endpoint to get user by ID

ehhong:
This is needed to support the org settings page and robot/location permissions.
Now that users can have robot/location authorizations, we do not want universally list all organization members. Instead, users should only be able to see other users that share authorizations on the resources they have access to.

To do so, we will fetch other users by the authorizations fetched from `listAuthorizations`, instead of relying on getting all org members put on `window.theData`

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -14,6 +14,9 @@ service AppService {
   // Get the id of the user with the email
   rpc GetUserIDByEmail(GetUserIDByEmailRequest) returns (GetUserIDByEmailResponse);
 
+  // Get user by ID
+  rpc GetUserByID(GetUserByIDRequest) returns (GetUserByIDResponse);
+
   // Organizations
 
   // Create a new organization
@@ -927,6 +930,21 @@ message GetUserIDByEmailRequest {
 
 message GetUserIDByEmailResponse {
   string user_id = 1;
+}
+
+message User {
+  string id = 1;
+  repeated string emails = 2;
+  google.protobuf.Timestamp date_added = 3;
+  google.protobuf.Timestamp last_login = 4;
+}
+
+message GetUserByIDRequest {
+  string id = 1;
+}
+
+message GetUserByIDResponse {
+  User user = 1;
 }
 
 message ListOrganizationsByUserRequest {


michaellee1019 (commented):
If I understand correctly, this will be used to lookup each user after we have made a ListAuthorizations call. To avoid the situations where X items in the list call results in x individual API calls, I would use [ListOrganizationMembers](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/app.proto#L41) instead. 

For permissions, I still think it is OK to know of all of the members of your org, but not know about the resources you don't have permissions to. I know that our UX doesn't have a section that lists org members without roles on resources. I think its still ok to do the filtering on the frontend? And use this API to eliminate the usage of window data?

I do think we need a GetUser API eventually, but would be more like a GetMyself API that returns information about the authenticated user.

ehhong:
> If I understand correctly, this will be used to lookup each user after we have made a ListAuthorizations call. To avoid the situations where X items in the list call results in x individual API calls, I would use [ListOrganizationMembers](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/app.proto#L41) instead.
> 
> For permissions, I still think it is OK to know of all of the members of your org, but not know about the resources you don't have permissions to. I know that our UX doesn't have a section that lists org members without roles on resources. I think its still ok to do the filtering on the frontend? And use this API to eliminate the usage of window data?
> 
> I do think we need a GetUser API eventually, but would be more like a GetMyself API that returns information about the authenticated user.

I think the specification from product states that we do not want to show all org members for users with only location/robot level permissions (row 35 of https://docs.google.com/spreadsheets/d/1G0Uw9FldVVqpFNV0TfkoloVvb8dhfRlPKYyGhKgGYTo/edit#gid=0)

Would the better move to add filtering to `ListOrganizationMembers`? Or like you said, do the filtering on the frontend

Kschappacher:
> > If I understand correctly, this will be used to lookup each user after we have made a ListAuthorizations call. To avoid the situations where X items in the list call results in x individual API calls, I would use [ListOrganizationMembers](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/app.proto#L41) instead.
> > For permissions, I still think it is OK to know of all of the members of your org, but not know about the resources you don't have permissions to. I know that our UX doesn't have a section that lists org members without roles on resources. I think its still ok to do the filtering on the frontend? And use this API to eliminate the usage of window data?
> > I do think we need a GetUser API eventually, but would be more like a GetMyself API that returns information about the authenticated user.
> 
> I think the specification from product states that we do not want to show all org members for users with only location/robot level permissions (row 35 of https://docs.google.com/spreadsheets/d/1G0Uw9FldVVqpFNV0TfkoloVvb8dhfRlPKYyGhKgGYTo/edit#gid=0)
> 
> Would the better move to add filtering to `ListOrganizationMembers`? Or like you said, do the filtering on the frontend

I like the filtering idea so we dont have a million backend calls. Frontend filtering would offload any work to the client but the data would still be sent to the user and technically they could still view it if they poked around more. If thats an issue I would do the filtering on the backend to ensure they only get data we want them to have

ehhong:
@Kschappacher @michaellee1019 Natalia confirmed that we should only show users that have authorizations on resources they can see.

We could add filtering to `ListOrganizationMembers`, but that would do a lot of duplicated work because we're already calling `ListAuthorizations` which is filtered. Another option to limit API calls is to call this endpoint `GetUsersByIDs`, but I understand if we don't want to advertise having functionality like that (even if the endpoint is permissioned)

From Natalia:
> Basically you see everyone added in your line of the tree.
> TLDR if a user can see the resource (even uneditable) then they see the people who have access to that resource.
image

michaellee1019:
Ok then these options should work:

1. Add same filtering to ListOrganizationMember. I hear the concern but can we refactor some things so that the two apis use the same filter call before returning responses (attempt to refactor so both use the same single like of permissions checking)

2. Add users to response of ListAuthorizations. This breaks the idea of different identity types, whcih clould be handled by either returning a user only for user type or create separate apis like ListUserAuthorizations and ListApiKeyAuthorizations, etc.

Still dont like this api because of the looping aspect of many calls when you have a predefined list already.

ehhong:
> Ok then these options should work:
> 
> 1. Add same filtering to ListOrganizationMember. I hear the concern but can we refactor some things so that the two apis use the same filter call before returning responses (attempt to refactor so both use the same single like of permissions checking)
> 2. Add users to response of ListAuthorizations. This breaks the idea of different identity types, which could be handled by either returning a user only for user type or create separate apis like ListUserAuthorizations and ListApiKeyAuthorizations, etc.
> 
> Still dont like this api because of the looping aspect of many calls when you have a predefined list already.

@michaellee1019 I like the idea of adding the same filtering to `ListOrganizationMembers` because I think it makes sense for that endpoint to be consistent with `ListLocations` and `ListRobots` (rather than only accessible by org operators / org owners). I also don't think it's clear that you'd want a bunch of misc user information with authorizations, seems like two separate concepts especially if more data is associated with a user

michaellee1019:
Ok lets do filtering in `ListOrganizationMembers` then!

ehhong:
closing this PR as this task no longer requires API changes

PR #316, opened 2023-08-02
DATA-1800 Add extra param to camera requests

kaywux:
Scope: https://docs.google.com/document/d/1-iYhuWE5umHaJ5H1v7SymLwEuKjy-12m-od_Rl6iXNY/edit


proto/viam/component/camera/v1/camera.proto
```diff
@@ -111,6 +114,8 @@ message RenderFrameRequest {
   string name = 1;
   // Requested MIME type of response
   string mime_type = 2;
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
```
tessavitabile, line 118: Not needed

kaywux: True - I added this for parity since [this line](https://github.com/viamrobotics/rdk/blob/6036fd070bc1d6228a293ae52c9f4748ca01a002/components/camera/server.go#L189) casts a RenderFrameRequest into a GetImageRequest, but can remove this change and instead construct an GetImageRequest with an empty extra field to reduce the additions we need to make to the api.

tessavitabile: Oops, sorry, I thought this was on GetImageResponse. This can stay as is, since we do want to have 'extra' in all requests.

kaywux: Sounds good!

tahiyasalam (approved):
LGTM :-)

PR #317, opened 2023-08-03
Add namespace to Module message

zaporter-work:
@ethanlook Mentioned that it would be easier for the frontend if we just provided the public namespace (if it exists) rather than making them parse the module id (or perform another lookup). Now that the namespace is attached to the module on the backend, this should be pretty easy to add on the backend.


proto/viam/app/v1/app.proto
```diff
@@ -877,6 +877,9 @@ message Module {
   string module_id = 1;
   // The id of the organization that owns the module
   string organization_id = 10;
+  // The public namespace of the organization that owns the module
+  // This is empty if no public namespace is set
+  string public_namespace = 12;
```
RoxyFarhad, line 882: nit: maybe add this so the numbering is correctly in order ?

RoxyFarhad (approved):
:)

proto/viam/app/v1/app.proto
zaporter-work: Add this so the numbering is in the correct order? There is already an `=11;` 

Do you just mean sort the entries in the message by the id?

RoxyFarhad: yeah just sorted in ID so its clear what the next value needs to be

michaellee1019 (approved):

proto/viam/app/v1/app.proto
zaporter-work: Sure -- I just left because it was already unsorted but happy to re-sort

ethanlook (approved):

PR #318, opened 2023-08-03
Add type to DeletePackageRequest

michaellee1019:
Currently DeletePackage does not accept a `PackageType` field. Internally if multiple packages exist with same id+version, the operation will fail due to more than one package existing. This follows the schema of Create/GetPackage to have the type be specified in the API request.


agreenb (approved):

RoxyFarhad (approved):
lgtm :)

PR #319, opened 2023-08-04
APP-2394 add ways to update and get CID

acrantel:
for use on the admin billing dashboard. the implementation on the backend will only include the field in the response or update the field if the user has the relevant permissions.


Kschappacher (approved):
LGTM

PR #325, opened 2023-08-10
APP-2028: Update organization member to include last login

anjinai:
**APP-2028: Update organization member to include last login**

## Change log

- Added an optional `last_login` field to the `OrganizationMember` protobuf definition.

### Notes

- We need to display the last_login date of each org member within the organization settings table in app.  This change will allow us to communicate `last_login` date and enable the frontend to display this data in the above mentioned table.

## Testing

- N/A

## Review requests

- None

[APP-2028]: https://viam.atlassian.net/browse/APP-2028


ehhong:
do we need to make this optional for backwards compatibility? change looks good to me, probably good to test locally before merging though just to make sure we don't break anything

anjinai:
> do we need to make this optional for backwards compatibility? change looks good to me, probably good to test locally before merging though just to make sure we don't break anything

I tested and it was fine/didn't break anything.

ehhong (approved):

PR #328, opened 2023-08-15
APP-2466 Update RPC API: CreateKey API protos, update Authorization proto

piokasar:
Adding a new rpc method CreateKey, as well as a corresponding message request and response. Adding a new field to Authorization message.

https://viam.atlassian.net/browse/APP-2466


ohEmily (approved):
LGTM! Thanks for doing this.

**I thought about it some more and wonder if before merging into `main`, maybe you could go ahead and create an `APP-2329` base branch on this repo and use that as the base branch instead (as I did in the app repo for [my first PR for this epic](https://github.com/viamrobotics/app/pull/2224).)**

My thinking is that if we are really trying to avoid breaking changes in this repo, it would be much safer to use a feature branch and should get into that habit. Maybe we discover new requirements, maybe we do end up making some minor changes. e.g. I really really hope we don't regret not calling this `CreateApiKey` instead of `CreateKey` and really hope  none of our existing APIs break after updating the Authorizations type, but hard to be sure without some manual testing.

piokasar:
Moved to[ a branch](https://github.com/viamrobotics/api/tree/app-2466-create-key-auth-update) on api. I'm going to close this PR.

PR #332, opened 2023-08-21
RSDK-4433 add package type for generic linux board definitions

oliviamiller:


penguinland (approved):
This is all we need!? That's pretty cool! LGTM

PR #333, opened 2023-08-23
APP-2536 - Cleanup unused orgid fields in module protos

zaporter-work:
Ticket : https://viam.atlassian.net/browse/APP-2536

Tested against https://github.com/viamrobotics/app/pull/2282
- No linting errors
- Make up succeeds
- No test failures (in the non-scheduler tests)

Will request review 1 week after https://github.com/viamrobotics/app/pull/2282 is merged to avoid breaking others who are using older branches off app.

Will re-test pre-merging.


zaporter-work:
Retested against app@head. No lint or test failures (on my machine).

michaellee1019 (approved):
LGTM!

zaporter-work:
Do either of you @DTCurrie  or @stevebriskin think that this requires a scope? These fields were never in the original scope and we thought they were needed but they never ended up being used.

DTCurrie:
> Do either of you @DTCurrie or @stevebriskin think that this requires a scope? These fields were never in the original scope and we thought they were needed but they never ended up being used.

If this was pre-release and never used I am ok with no scope doc, but I think @stevebriskin will have to make the final call since he has the power to merge.

michaellee1019:
> I am ok with no scope doc

These changes were already included in a email thread for the updated scope (amongst many other things). It was sent a few weeks ago but didn't do the PR until now.

PR #334, opened 2023-08-23
APP-2543 - Attach version info to returned objects from GetModule

zaporter-work:
Ticket : https://viam.atlassian.net/browse/APP-2543

This is important for upload validation.

Right now I am not including description and URL because not sure if this is the best way to expose that info. Happy to add those if either of you think it would be good to have.


RoxyFarhad (approved):
lgtm!

michaellee1019 (approved):
I think we should cleanup the `Module` message. Not right now. Lets create a task and do it when appropriate?

zaporter-work:
Good idea. Created a ticket that we can triage later https://viam.atlassian.net/browse/APP-2558

PR #335, opened 2023-08-24
APP-2530 Update OrganizationInvite message

ehhong:
### Motivation
We need to move organization invites off of `window.theData` and instead fetch it from an endpoint to support the Svelte migration. The endpoint `ListOrganizationMembers` already returns invites, but the `OrganizationInvite` message has not been updated with `authorizations`. 

### Changes
- additive
     - add `authorizations`
- breaking
     - remove `robot_count` (accidentally added)


stevebriskin (approved):

DTCurrie (approved):
✅

PR #336, opened 2023-08-24
APP-2545: Add key_id UUID to API keys and randomly generate key

ohEmily:
[APP-2545](https://viam.atlassian.net/browse/APP-2545)

Adds the new `key_id` field to `CreateKeyResponse`. This addresses Emily's comment [here](https://docs.google.com/document/d/1ovx8AKwxobkO2860vLKHPpbKexNrmepYVXsoLYwEg60/edit?disco=AAAA4WZWBzk) about not currently supporting authentication against resources other than the resource the key was created against. By adding a `key_id`, we can query against the `key_id` while passing the secret instead of against the specific resource.

[APP-2545]: https://viam.atlassian.net/browse/APP-2545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ehhong:
what do you think of calling it `id` and `key`? (instead of `key_id` and `key`)

ehhong (approved):
one open question about naming, otherwise lgtm!

ohEmily:
WFM!

ehhong (approved):

PR #337, opened 2023-08-25
Add Movement sensor units to proto

randhid:


proto/viam/component/movementsensor/v1/movementsensor.proto
```diff
@@ -117,6 +117,8 @@ message GetOrientationRequest {
 }
 
 message GetOrientationResponse {
+  // Orientation is returned as an orientation message with 
+  // OX OY OZ dimensionalles and Theta in degrees
```
jeremyrhyde, line 121: [opt] 
```suggestion
  // OX OY OZ and Theta in degrees
```

jeremyrhyde (approved):
LGTM, second option for one of the comments but is a nit/opt

proto/viam/component/movementsensor/v1/movementsensor.proto
randhid: This isn't true, but I misspelled dimensionless.

PR #338, opened 2023-08-25
APP-2543 - Fix typo

zaporter-work:
This typo was introduced by me in https://github.com/viamrobotics/api/pull/334 two hours ago. It is very unlikely that anyone is using this version of API yet. 

I read over that pr 2 times but it still slipped by. I will read more carefully going forward.

If it is easier, we can revert that PR and I can re-do it or we can fail-forward with this PR.


michaellee1019 (approved):
Will require asking steve to merge in.

stevebriskin (approved):

PR #344, opened 2023-09-05
[DATA-1909] Add any type to sensor and tabular data messages

vijayvuyyuru:
This change is part of queryable tabular data. I updated the `TabularData` and `SensorData` message to include this new any field as discussed [here](https://docs.google.com/document/d/1cdNFkIwqNAMaev4jvyr0r6tH-4qOS51vgJsWGIMdmBo/edit). The current struct was not removed from either to keep migration/coordinating this work easier.


vijayvuyyuru:
Changing approach to sensorData api

PR #346, opened 2023-09-07
APP-2589: Add support for DeleteKey and ListKeys endpoints

jr22:
[APP-2589](https://viam.atlassian.net/browse/APP-2589)

* Created two new endpoints to delete an api key (DeleteKey) and to list all the keys associated with a given resource (ListKeys) 
* Implemented in app side PR https://github.com/viamrobotics/app/pull/2420/files#diff-98a62a7ad1111c58a8b0375abccd0cd94f7d7b6de44dbb602ed29614de41eb4bR1852-R1855
* Will remove compiled changes before merging but added them in so I could do local testing with my app branch 

[APP-2589]: https://viam.atlassian.net/browse/APP-2589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


proto/viam/app/v1/app.proto
```diff
@@ -975,3 +979,24 @@ message CreateKeyResponse {
   string key = 1;
   string id = 2;
 }
+
+message DeleteKeyRequest {
+  string key = 1;
+  string id = 2;
+}
+
+message DeleteKeyResponse {}
```
jr22, line 988: noticed that the pattern seems to still declare a response type even if its empty. is that correct? very new to proto stuff

ehhong: i think that's correct!

proto/viam/app/v1/app.proto
```diff
@@ -975,3 +979,23 @@ message CreateKeyResponse {
   string key = 1;
   string id = 2;
 }
+
+message DeleteKeyRequest {
+  string key = 1;
+  string id = 2;
```
ehhong, line 985: is it typical to request both the id and key to delete an API key? this looks good to me but I just wanted to double check to see that this is the usual pattern

proto/viam/app/v1/app.proto
```diff
@@ -975,3 +979,23 @@ message CreateKeyResponse {
...
+
+message DeleteKeyResponse {}
+
+message ApiKey {
+  string key = 1;
+  string id = 2;
+}
+
+message ListKeysRequest {
+  string org_id = 1;
```
RoxyFarhad, line 996: just noticed this while looking at your PR for other reasons. Is the reason you are passing this in for if we are trying to list keys from somewhere that is not the APP? 

the org_id should be passed in within the context object so if this is from APP you probably don't need to explicitly pass this in the request.

ehhong: ah actually we are trying to remove support for current org in the BE so we should still explicitly include it here

RoxyFarhad: ah okay wasn't aware of that -- good to know thanks!

jr22: for my own learnings, is there anything else we include in the context automatically and where in the code does this happen?

proto/viam/app/v1/app.proto
```diff
@@ -975,3 +979,23 @@ message CreateKeyResponse {
...
 }
+
+message DeleteKeyRequest {
+  string key = 1;
+  string id = 2;
+}
+
+message DeleteKeyResponse {}
+
+message ApiKey {
```
RoxyFarhad, line 990: @jr22 am going to copy and paste this into my current API Pr as we both seem to need it. Shouldn't be an issue to resolve for whoever gets there first :)

proto/viam/app/v1/app.proto
ehhong: i think it's everything that populates the state object here, it's mostly authentication/authorization information. except `currentOrg` lol, which we are trying to deprecate because it is unreliable and is only relevant for frontend state
https://github.com/viamrobotics/app/blob/0c1b93bf302d9ef4d70b9dd4995da8162622c15e/auth/state.go#L21-L32

jr22: got it! thank you!!

proto/viam/app/v1/app.proto
jr22: also seeing EPs comment on your PR, reminded me that I actually probably need to include some more metadata here as well for the orgs page (authorizations etc)

jr22: So on the frontend we would do something like: call `ListKeys` immediately followed by call to `ListAuthorizations` to get the details? Is that what you mean? 

If I'm understanding the code correctly, `ListOrganizationMembers` seems like it is querying the organization collection (mostly, although i do see the $lookups) and then `ListAuthorizations` queries the `identity authorizations` collection. In this case, I think its slightly different since we would be querying the `main.identity_authorization` collection for _both_ endpoints and for `ListKeys` we actually already have all the authorization information included in the db response, the current code is just filtering it out (due to slight misunderstanding by me for how this was going to be used). Given that it would be a duplicate db call, I _think_ it makes the most sense to include all the available info in the `ListKeys` response, but def open to push back since I'm still learning about the system!

proto/viam/app/v1/app.proto
jr22: Yeah I think this is related to the comments on Roxy's related PRs and is explained in the spec for org api keys (under "Why do we need both a key and an ID field for the ID? Isn't a single cryptographically generated key already safe?"). Let me know if you still have questions!

proto/viam/app/v1/app.proto
ehhong: ah that's a fair point! also we probably will never need to return a key without its authorizations

jr22: continuing convo from roxys pr here. updated mine to have one field that leverages APIKey and one that lists the authorizations. i'm not sooo in love with the key details being a separate field on the object from authorizations (feels like it implies that the authorizations aren't part of the key.. which they are) but also feel like this is the only way to get it to leverage the APIKey and avoid duplication. curious to hear what you guys think.

ehhong: i think it's a fine split because there is a valid use case for why you'd want just the API key and not its authorizations (ex. code samples). i feel like a user is pretty analogous in this situation, where it feels pretty natural that a user exists independently of its authorizations.

proto/viam/app/v1/app.proto
```diff
@@ -987,3 +991,31 @@ message CreateKeyResponse {
...
+
+message DeleteKeyResponse {}
+
+message AuthorizationDetails {
+  string authorization_type = 1;
+  string authorization_id = 2;
+  string resource_type = 3;
+  string resource_id = 4;
+  string org_id = 5;
+}
```
ehhong, line 1008: [q] why did you decide to create a new message instead of reusing the existing `Authorization` message? was it to avoid the redundant fields (identity_id, identity_type, etc)?

proto/viam/app/v1/app.proto
RoxyFarhad: I see what you mean here about the nesting, but yeah there is an argument that it also actually matches the DB object as that is nested. I would argue duplicating a few fields isn't that big of a deal,so I would favor just repeating the API key fields here and then adding the nested object for authorizations.

proto/viam/app/v1/app.proto
jr22: Yeah I thought it made more logical sense with how the information will likely be used. The additional duplicated information can be a bit confusing I think

ehhong (approved):
looks good to me! pending discussion on whether these endpoints should request the API key

RoxyFarhad (approved):
lgtm too :)

proto/viam/app/v1/app.proto
```diff
@@ -987,3 +991,33 @@ message CreateKeyResponse {
...
+
+message APIKeyWithAuthorizations {
+  string id = 1;
+  string key = 2;
+  string name = 3;
+  google.protobuf.Timestamp created_on = 4;
```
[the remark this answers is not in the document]
jr22: i slightly prefer the flat structure but would also love to keep this ticket moving and don't feel strongly enough so changed it back. hopefully developing some more concrete guidelines on this in the future will help

ehhong (approved):
lgtm!! thanks for all the patience on the back and forth here

PR #351, opened 2023-09-11
RSDK-4199 - Remove pbstruct from mlmodel.proto

kharijarrett:


kharijarrett:
The only actual _change_ here is that I'm removing the pb.Struct as a potential input for the Infer Request/Response. It's still around for potential "extra".

bhaney (approved):
LGTM

proto/viam/service/mlmodel/v1/mlmodel.proto
```diff
@@ -28,18 +28,14 @@ service MLModelService {
 message InferRequest {
   // name of the model service
   string name = 1;
-  // this is a struct of input arrays/tensors as specified in the metadata
-  google.protobuf.Struct input_data = 2;
-  // the input data can also be provided as set of named flat tensors
+  // the input data is provided as set of named flat tensors
```
tahiyasalam, line 31: I think we should try to reserve the proto field after deleting more info [here](https://protobuf.dev/programming-guides/proto3/#deleting). Otherwise, the user can accidentally use this field and break the existing code. There are some example [here](https://github.com/viamrobotics/api/blob/main/proto/viam/app/data/v1/data.proto#L104) of deleting reserved fields.

tahiyasalam (changes requested):
This looks great! Super excited about this change. A nit on my end to reserve the field after deleting

proto/viam/service/mlmodel/v1/mlmodel.proto
kharijarrett: Good call! Just added the reserved fields!

tahiyasalam (approved):
LGTM, thanks for adding me to this !

Otterverse (approved):
Kick the CI...

Otterverse (approved):
Go go gadget CI!

PR #353, opened 2023-09-12
APP-2625: getRobotAPIKeys request

RoxyFarhad:
For this ticket: https://viam.atlassian.net/browse/APP-2625


ohEmily (changes requested):
Small oops here -- API key auth requires key ID and the key itself at the very least for authentication. Assuming this endpoint is for the code samples page only, we would need to return both of those.

I think I support having the code samples have their own endpoint for simplicity's sake? Otherwise we would need several optional parameters for loading API keys to support the eventual full UI shown [here](https://www.figma.com/file/2hs9zGWN9nTLqcm8pR38ip/Management?type=design&node-id=4987-3182&mode=design&t=m7tF2NzRQMBiKEXT-0) (this is the in-progress figma file attached to the robot & location epic).

RoxyFarhad:
@ohEmily 
> API key auth requires key ID and the key itself at the very least for authentication. Assuming this endpoint is for the code samples page only, we would need to return both of those.

Ah so we want to return a list of the Key ID, key value pairs? for my own understanding, why is the key ID required for authorization? My assumption was that if you have the api-key, that is enough for authorization?

jr22:
FYI adding something similar in my PR here: 
https://github.com/viamrobotics/api/pull/346

Think this will be used to drive the UI that EP linked in her comment. If we want to keep them separate, think we're fine but we can also chat to make sure we dont step on each other's toes too much :D

ohEmily:
Hey, sorry I just stepped away from my desk but if you take a look at the org API key spec it shows what the auth schema looks like for keys and there's a Q&A section explaining why we have both keys and key IDs!

proto/viam/app/v1/app.proto
```diff
@@ -624,6 +627,19 @@ message DeleteRobotPartRequest {
...
 }
 
+message GetRobotAPIKeysRequest {
+  string robot_id = 1;
+}
+
+message ApiKey {
+  string key = 1;
+  string id = 2;
+}
```
ohEmily, line 637: ah ok if we're adding an object, should we include the optional fields in the schema? e.g. `created_by`, `name` etc. or should we have a separate protobuf definition for those APIs?

RoxyFarhad: We can add created_by, name, and robot/location resource count as well. I had issues loading the figma design till now, but now I can see those fields. All of those except for name can be populated now, but we can just return "name" as an empty string till then. When are we planning to add the name field? as I don't see it as a field in the DB entry for IdentityAuth and I can't find the ticket

proto/viam/app/v1/app.proto
```diff
@@ -624,6 +627,23 @@ message DeleteRobotPartRequest {
...
+message GetRobotAPIKeysRequest {
+  string robot_id = 1;
+}
+
+message ApiKey {
+  string id = 1;
+  string key = 2;
+  string name = 3;
+  string created_on = 4; 
+  repeated string locations = 5;
```
RoxyFarhad, line 639: do we want more information for what resources are using this key apart from a list of names? Could potentially see IDs as helpful if we want to link to a certain page down the line, but that may be too out of scope

also @jr22 if you think this will conflict with your version of ApiKey, I can make some fields optional or change the name.

jr22: cool thanks! i think for mine ill need similar fields (name, createdOn etc) but I'll also need to include some metadata about the resources (i.e. in that table how it says an api key is on 5 locations and 2 robots for example). not sure if you need that for this api. is this `GetRobotAPIKeys` request primary purpose to be called by the code sample page?

RoxyFarhad: yup you are right, i got confused and looked at the wrong designs. Changed this now to be "RobotAPIKey" so we can diff these out. My bad for the confusion sorry !

proto/viam/app/v1/app.proto
```diff
@@ -624,6 +627,21 @@ message DeleteRobotPartRequest {
   string part_id = 1;
 }
 
+message GetRobotAPIKeysRequest {
+  string robot_id = 1;
+}
+
+message RobotApiKey {
```
ohEmily, line 634: [minor] capitalization doesn't match othoer stuff in this PR, this should be `RobotAPIKey`

ohEmily (approved):
LGTM once capitalization is fixed. Probably makes more sense for Julia to review than me since her PR will be consuming this interface now

proto/viam/app/v1/app.proto
jr22: Might be worth calling this `APIKey` instead of `RobotAPIKey` so then i can add the optional `AuthorizationDetails` field in the response and reuse it. Thoughts?

RoxyFarhad: yeah either works for me -- makes sense to converge as they map to the same object

jr22 (approved):
LGTM :shipit:

proto/viam/app/v1/app.proto
ehhong: hmm if authorizations are optional i think it makes it more difficult to know what to expect from each endpoint. i think `optional` should mean "will be returned if the value is available", not "will be returned depending on the context"

maybe this is a comment for APP-2589, but what do you think of having one message `APIKey` with the `key`, `id`, `created_on`, etc. and then returning the authorizations as a separate message? that way we can reuse `APIKey` here, because I agree it doesn't really make sense to have `RobotAPIKey` and `APIKey`. if we want two separate messages it should probably be something like `APIKey` and `APIKeyWithAuthorizations`

jr22: sure yeah i think thats valid. I think for Roxy's PR it probably makes sense to leave as is and then we can chat and fix on mine. sound good?

ehhong: the only thing relevant to Roxy's PR is the name of the message. otherwise you might have to make a breaking change on yours. how do you guys feel about `RobotAPIKey` versus `APIKey`? i am leaning towards `APIKey` because there is nothing in this proto message that is particular to robot API keys imo

jr22: oh yeah when i said "leave as is" i meant with `APIKey`. I think `APIKey` makes the most sense as well.

ehhong (approved):

PR #358, opened 2023-09-20
Added proto messages for adding a role to and removing role from api key

Oabraham1:
I will be adding the `api_key` field to the Authorization type definition to enable us more easily support `AddRole` and `RemoveRole` for api keys. Adding the API Key field to the authorization field enables us perform necessary lookups for during authorization.

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -773,6 +773,7 @@ message Authorization {
   string identity_id = 5;
   string organization_id = 6;
   string identity_type = 7;
+  string api_key = 8;
 }
 
 message AddRoleRequest {


Oabraham1:
Thinking about this even more and I don't think we even need to add the API Key to the authorization since we could just pass in the api key id as the identityID and use that to perform necessary lookups. Let me know what you think @ehhong @piokasar

ehhong:
> Thinking about this even more and I don't think we even need to add the API Key to the authorization since we could just pass in the api key id as the identityID and use that to perform necessary lookups.

I agree with this!

Oabraham1:
> > Thinking about this even more and I don't think we even need to add the API Key to the authorization since we could just pass in the api key id as the identityID and use that to perform necessary lookups.
> 
> I agree with this!

Thanks Emily, gonna work on refactoring the code to test out this hypotheses.

Oabraham1:
Closing this PR since we're not using api key in the authorization request anymore

PR #359, opened 2023-09-20
[DATA-1861] Remove DeleteTabularDataByFilter from API

agreenb:
Both stable RDK and Python SDK are no longer relying on DeleteTabularDataByFilter, and instead using DeleteTabularData.

Removed all implementation and remaining app use of it in https://github.com/viamrobotics/app/pull/2507


tessavitabile:
Should I go ahead and merge?

agreenb:
@tessavitabile yep you can merge whenever, thank you!

PR #361, opened 2023-09-21
Rotate key

Kschappacher:
add rotate key proto methods


proto/viam/app/v1/app.proto
```diff
@@ -1018,3 +1021,13 @@ message ListKeysRequest {
 message ListKeysResponse {
   repeated APIKeyWithAuthorizations api_keys = 1;
 }
+
+message RotateKeyRequest {
+  string key = 1;
+  string id =2;
+}
```
jr22, line 1028: so i think since we can remove `key` from this request since I think we decided you don't need the `key` to be able to do operations like rotate and delete. i do realize though that this will mess up the call to `IdentityAuthorizationGetForAPIKey` and [this call](https://github.com/viamrobotics/app/pull/2499/files#diff-5d4eeb9a95bb6dbd603dec16eb0fa3e6a5367a6bfa2e2379ee411c777afbfbb7R1951) until [APP-2776](https://viam.atlassian.net/browse/APP-2776) is done though. what i did to get around this for deleteKey is something like this: 
https://github.com/viamrobotics/app/blob/main/auth/permissioned_data.go#L548
https://github.com/viamrobotics/app/blob/main/data/data.go#L1948-L1951

Otterverse:
Just a note. I see you manually added the protos-compiled label. You should never do that. The workflow adds that (after approval) and after it's actually compiled and merged the generated protos. I'm removing it, and once you get actual approval, it should add it back automatically.

Kschappacher:
> Just a note. I see you manually added the protos-compiled label. You should never do that. The workflow adds that (after approval) and after it's actually compiled and merged the generated protos. I'm removing it, and once you get actual approval, it should add it back automatically.

Thanks for the heads up! havent had to make changes to api for a while and musta forgotten

proto/viam/app/v1/app.proto
Kschappacher: that makes sense. just pushed the change through

proto/viam/app/v1/app.proto
```diff
@@ -1018,3 +1021,11 @@ message ListKeysRequest {
...
   repeated APIKeyWithAuthorizations api_keys = 1;
 }
+
+message RotateKeyRequest {
+  string id = 1;
+}
+
+message RotateKeyResponse {
+  string id = 1;
+}
```
jr22, line 1031: oh sorry i wasn't super clear in my original comment but i DO think we want to include `key` in the response. since we're generating a new key, we'll want to send that along to display. although i haven't fully though through the how the client side will be calling this. do you think the UI will make this rotate key request and replace what's shown in the api keys table? or do you think the UI should recall `ListKeys` after rotating?

Otterverse:
Sorry to jump in here again, but you're not supposed to make any further code changes after approval, because the compile workflow already ran. Looks like you merged stuff afterwards, so you'll need to revert a lot of things and do it again. Basically, manually remove the 'protos-compiled' label, remove the generated files from the PR branch, (so files-changed shows only the .proto files) and then have approval resubmitted. After that it should generate things based on all the changes together.

The steps are outlined in the readme here for future reference and/or troubleshooting: https://github.com/viamrobotics/api#submitting-prs-for-protobuf-changes

PR #363, opened 2023-09-25
DATA-1982 - Remove AddBinaryDataToDatasetByFilter in API

tahiyasalam:
We're no longer implementing this


tessavitabile (approved):
Is this ready for me to merge?

tahiyasalam:
@tessavitabile Yep! This is not used anywhere and was unimplemented, so this should not break anything :)

PR #364, opened 2023-09-25
RSDK-4786-dry-proposal

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-4786)

## Reason for Change: 
When we started implementing the GRPC stubs for the new motion methods we discovered that GetPlanResponse duplicated the `plan_id`, `execution_id`, and `component_name` in the `GetPlanResponse` between both the `Plan` and `PlanStatus` fields. 

We believe this duplication will make it ambiguous for API callers, which subfields (`Plan` or `PlanStatus`) callers should consider the source of truth.

This was a consequence of having only a single type for returning status information (`PlanStatus`) between the `ListPlanStatusesResponse` (which needs to include `plan_id`, `execution_id`, and `component_name` )   and `GetPlanResponse` (which does not need to include `plan_id`, `execution_id`, and `component_name`, as those fields are already present on the `Plan)`.

This change DRYs up the API by creating a new time called `PlanStatusWithID` which contains the  `plan_id`, `component_name` and `execution_id` fields &  removes those fields from `PlanStatus`. 

## Changes:
1. Change `motion.ListPlanStatusesResponse.statuses`'s type from `PlanStatus` to `PlanStatusWithID` & rename it to `status_summaries`
2. Remove `plan_id`, `component_name` and `execution_id` from `PlanStatus`
3. Create `PlanStatusWithID`, which contains  `plan_id`, `component_name` and `execution_id`  & a `PlanStatus`


tessavitabile (approved):
LGTM. Since this is a breaking change in the rdk API, I will merge this on Thursday morning prior to when we cut a release, to avoid mid-release-cycle incompatibility between app and the previous rdk.

proto/viam/service/motion/v1/motion.proto
```diff
@@ -207,8 +207,8 @@ message StopPlanRequest {
   // The name of the motion service
   string name = 1;
 
-  // The root component of the currently executing plan to stop
-  common.v1.ResourceName root_component = 2;
+  // The component of the currently executing plan to stop
+  common.v1.ResourceName component_name = 2;
```
nicksanford, line 211: @raybjork & @tessavitabile I'd like to also add this change in this PR. Calling this field `root_component` in the `StopPlanRequest` was a typo from a previous iteration of the scope doc. Nowhere else uses the term `root_component` & eliot specifically asked us to use `component_name` everywhere we send a component name.

Can we please roll this fix into this PR?

raybjork: Thats good with me

tessavitabile: same

PR #379, opened 2023-10-09
RSDK-4787: Add Readings to API

martha-johnston:
Move GetReadingsRequest and GetReadingsResponse to common.proto and add that method to sensor.proto movementsensor.proto and powersensor.proto.


proto/viam/component/movementsensor/v1/movementsensor.proto
```diff
@@ -70,6 +70,13 @@ service MovementSensorService {
       get: "/viam/api/v1/component/movementsensor/{name}/geometries"
     };
   }
+
+  // GetReadings returns the readings of a sensor of the underlying robot.
+  rpc GetReadings(common.v1.GetReadingsRequest) returns (common.v1.GetReadingsResponse) {
+    option (google.api.http) = {
+      get: "/viam/api/v1/component/sensor/{name}/readings"
```
oliviamiller, line 77: should be `movementsensor` instead of `sensor ` and same in power sensor

oliviamiller (approved):
one nit then lgtm! I think you'll have to ask an admin to push since its a breaking change

randhid:
@stevebriskin for breaking changes that won't really break anything/

stevebriskin:
LGTM
This is not a breaking wire protocol change (package renaming is backward compatible over the wire). Old SDKs will continue to work with the new RDK and vice versa.
This change will break builds for everyone relying on compiled proto files directly (e.g. all SDKs that have wrappers).
This change will break all users of raw compiled proto structs. Also impacts SDKs that don't have wrappers (e.g. C++)

PR #381, opened 2023-10-11
APP-2909 - Update API to have new env variable fields

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2909

Scope: https://docs.google.com/document/d/1N2U2dW2TtP-R89VPhYyJdAvaeZzogA2t3_Vc2VlS0to/edit

@benjirewis added for visibility


proto/viam/app/v1/robot.proto
```diff
@@ -326,6 +328,12 @@ message ModuleConfig {
   string path = 2;
   // log level for module
   string log_level = 3;
+  // type of the module ("local" or "registry")
+  string type = 4;
+  // the id of the module if it is a registry module
+  string module_id = 5;
```
zaporter-work, line 334: The scope has this as optional but it didn't fit our current paradigm for what is optional / required and made the proto conversions confusing.

I am biasing towards consistency with the rest of this message (and other similar messages)

RoxyFarhad (approved):
:)

proto/viam/app/v1/robot.proto
benjirewis: > it didn't fit our current paradigm for what is optional / required and made the proto conversions confusing.

What is our paradigm of optional vs required for fields that are not always specified? I assume `module_id` will be `""` if you don't specify it? I guess we only need `optional` if there's some distinction between `""` and not specified?

benjirewis (approved):

proto/viam/app/v1/robot.proto
zaporter-work: Yeah, module_id would be "". Ex: log_level is optional but we use an empty string to denote that it is unspecified rather than an optional param.

PR #383, opened 2023-10-13
APP-2888: add AppValidationStatus field for api

RoxyFarhad:
If there are cloud validations that fail on config retrieval, we want to return that the user but not actually return an error. This accomplishes that. 

Goes hand-in-hand with: 
https://github.com/viamrobotics/app/pull/2672 
https://github.com/viamrobotics/rdk/pull/3078

- this needs to be merged in first 
- and then RDK + then APP

### DIFF
--- a/proto/viam/app/v1/robot.proto
+++ b/proto/viam/app/v1/robot.proto
@@ -47,6 +47,10 @@ message LocationSecret {
   string secret = 2;
 }
 
+message AppValidationStatus {
+  string error = 1;
+}
+
 message CloudConfig {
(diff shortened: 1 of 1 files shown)


michaellee1019 (approved):

michaellee1019:
@stevebriskin does this require a scope?

stevebriskin:
Yes. The new flow is pretty unclear to me and this new field feels like a pretty hacky addition to the most critical rdk<->app api.

michaellee1019 (approved):

PR #385, opened 2023-10-24
APP-1865: remove all the old billing protos that are not supported

RoxyFarhad:
Were never used these protos as we changed the design of the rpc calls halfway through billing project and no longer need these. I tested that this doesn't break anything in APP locally, but @stevebriskin I think its technically a breaking change so you need to force merge.


ohEmily:
- need to merge main back in
 - add protos compiled
 - then tag Steve in to hit merge

RoxyFarhad:
Tested locally with latest main pointing to this version of API and nothing breaks.

PR #386, opened 2023-10-26
[APP-2783]: (API) Simplify shared location permissions

maxhorowitz:
This is a branch containing the API changes required in the Simplify shared location permissions project. Reference [this PR](https://github.com/viamrobotics/app/pull/2783) for the entire project.


proto/viam/app/v1/app.proto
```diff
@@ -371,8 +371,10 @@ message DeleteOrganizationMemberResponse {}
 message LocationOrganization {
   // Organization ID the location is shared with.
   string organization_id = 1;
+  // Organiztion name the location is shared with.
+  string organization_name = 2;
   // Whether the organization is the primary owner or not.
-  bool primary = 2;
+  bool primary = 3;
```
ohEmily, line 377: oops this GitHub Action failure detection is legitimate! We can't reorder the fields in a proto. We'll have to make `organization_name` set to 3. See the [official docs](https://protobuf.dev/programming-guides/proto3/#assigning) for more on assigning field numbers.

proto/viam/app/v1/app.proto
```diff
@@ -371,8 +371,10 @@ message DeleteOrganizationMemberResponse {}
 message LocationOrganization {
   // Organization ID the location is shared with.
   string organization_id = 1;
+  // Organiztion name the location is shared with.
+  string organization_name = 2;
```
ohEmily, line 375: I don't quite understand why this field is required. If i'm not mistaken, the [app PR](https://github.com/viamrobotics/app/pull/2780/files#diff-515b343b0aed1bc7f56fc1d71a890e243bc0e3f154c7edd8d101193997083a7c) doesn't use this updated copy of `api` yet so I wasn't able to find your utilization of the field. Let me know if there's somewhere else I should be looking!

ohEmily (changes requested):
Hitting request changes since we have a breaking change here (reusing a field number) but looking forward to hearing more about the `organization_name` field. Thank you for organizing your PRs so clearly with links between them!

PR #388, opened 2023-10-26
RSDK-4567 add gostream proto

bazile-clyde:
# Description
See https://github.com/viamrobotics/rdk/pull/3000#discussion_r1367996614

This copies the proto file from gostream to the API. This does _not_ copy any of the generated `*.pb` files. That seems to be the standard but let me know if that isn't the case.

# Testing
I tested this locally by replacing the API dependency in the RDK with a local copy of the API with these changes and ran `make test-go` along with the viam-server.


Otterverse (approved):
LGTM! Now just wait a few minutes for it to run the generation workflow and update the PR before you merge.

PR #391, opened 2023-10-27
Bump golang.org/x version to 0.17.0

stevebriskin:


Otterverse (approved):
LGTM

Otterverse:
FYI, due to the bug in the workflows, I'm having to manually add a random tag (I use "enhancement") after the protos are compiled in order for it to pass the checks.

stevebriskin:
> FYI, due to the bug in the workflows, I'm having to manually add a random tag (I use "enhancement") after the protos are compiled in order for it to pass the checks.

tx. i wasn't sure what was happening.

PR #394, opened 2023-11-06
DATA-1109 - Add training start time to metadata

tahiyasalam:
Changes staged [here](https://github.com/viamrobotics/app/pull/2855). I tested everything locally, but I will merge, update API and `Deploy` once this is merged.


agreenb (approved):
LGTM!

Can you check that this doesn't cause any linter errors in app? It shouldn't since we're not creating a TrainingJobMetadata (and if we were, the linter would complain that some fields are not explicitly set) and instead just passing it through. Can check this locally with app pointing to this PR and running `cd frontend && npm run lint`

PR #395, opened 2023-11-09
DATA-2104 - Add TabularDataBySQL and TabularDataByMQL APIs

dmhilly:
This PR adds two new data management APIs, TabularDataBySQL and TabularDataByMQL. Users can query their tabular data using SQL or MQL using these APIs.


agreenb (approved):
💯

PR #404, opened 2023-11-27
APP-3181 ListRegistryItem

ehhong:
API changes to support ListRegistryItem. Includes proto definition for RegistryItem


proto/viam/app/v1/app.proto
```diff
@@ -830,6 +832,41 @@ message CheckPermissionsResponse {
...
+  // The url to reference for documentation, code, etc.
+  string url = 8;
+  // A short description of the item that explains its purpose
+  string description = 9;
+
+  //For Modules Only
+  // A list of models that are available in the module
+  repeated Model models = 10;
+  // The executable to run to start the module program
+  optional string entrypoint = 11;
```
ehhong, line 862: changed this from the [scope](https://docs.google.com/document/d/1-eEz5r2hJL4WaUIVoNAWKKPSTPe6k3VuqUyJP11IXo8/edit) to be optional (since it is for modules only) -- does that seem correct?

proto/viam/app/v1/app.proto
```diff
@@ -830,6 +832,57 @@ message CheckPermissionsResponse {
   repeated AuthorizedPermissions authorized_permissions = 1;
 }
 
+message ModuleVersion {
```
ehhong, line 835: made a new message (that is essentially a duplicate of `VersionHistory`) to make this a little more flexible. 
is it possible to deprecate `VersionHistory` when modules are deprecated? if not i will reuse `VersionHistory`

zaporter-work: I think this is a good move long term. This is named better than `VersionHistory`.

What do you think about the name `ModuleVersionMetadata` to clearly indicate that it belongs as a subtype of module metadata?

proto/viam/app/v1/app.proto
```diff
@@ -830,6 +832,57 @@ message CheckPermissionsResponse {
...
+  repeated Model models = 3;
+  // The entrypoint for this version of the module
+  string entrypoint = 4;
+}
+
+message ModuleMetadata {
+  // A list of models that are available in the module
+  repeated Model models = 1;
+  // A list of versions of the module that are available
+  repeated ModuleVersion versions = 2;
```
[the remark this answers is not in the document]
zaporter-work: To solidify our offline discussion, if I am understanding this right, when you add ml models to the registry, you would add something like:
```
message MlModelMetadata {
   ....
    repeated MlModelVersionMetadata versions = xyz;
}
```
right?

(because I seem to recall the data team talking about how not all ml_model packages should turn into registry versions so you need metadata about which models actually ended up in the registry)

proto/viam/app/v1/app.proto
```diff
@@ -850,6 +903,27 @@ message UpdateRegistryItemRequest {
 
 message UpdateRegistryItemResponse {}
 
+enum RegistryItemStatus {
+  REGISTRY_ITEM_STATUS_UNSPECIFIED = 0;
+  REGISTRY_ITEM_STATUS_PUBLISHED = 1;
+  REGISTRY_ITEM_STATUS_IN_DEVELOPMENT = 2;
+}
+
+message ListRegistryItemsRequest {
```
zaporter-work, line 912: [opt/pedantic/might be too late to change] Should this be something like `SearchRegistryItemsRequest`.

It seems like this is more than just a basic listing

proto/viam/app/v1/app.proto
```diff
@@ -830,6 +832,57 @@ message CheckPermissionsResponse {
...
+  // The models that this verion of the module provides
+  repeated Model models = 3;
+  // The entrypoint for this version of the module
+  string entrypoint = 4;
+}
+
+message ModuleMetadata {
+  // A list of models that are available in the module
+  repeated Model models = 1;
+  // A list of versions of the module that are available
```
zaporter-work, line 849: [nit] can you include the comment about the sort order that was on the module message or the version message or somewhere

zaporter-work (commented):
LGTM! (not approving so that protos don't generate before you are ready)

proto/viam/app/v1/app.proto
ehhong: i like `ModuleVersion` because it is clear that it represents a version of a module (whereas it's not clear what "VersionMetadata" is). I think we can rely on docs / product communication for a user to understand that module versions are within a registry's module metadata

proto/viam/app/v1/app.proto
ehhong: ah agreed that is maybe a better name, but at this point all of our "return multiple items of this type" functions are called "List<item>" (see "ListAuthorizations", "ListLocations", "ListRobots" etc). it's not too late to change but i'm inclined to leave it as is

ehhong: but let me know if you feel strongly -- could be convinced that this is a different enough use case that it warrants "search"

proto/viam/app/v1/app.proto
michaellee1019: Naming makes sense to me, the version fields here are specific to modules.

proto/viam/app/v1/app.proto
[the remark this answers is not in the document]
michaellee1019: This was my main thought on this PR. Any reason we can't add the ML Model versioning now? We know we need it and there won't be a way to show versions for ML Models without it (except for calling ListPackages which we don't want to do).

proto/viam/app/v1/app.proto
michaellee1019: Lets stick with List because it matches our conventions. A lot of List APIs have optional filters on them.

proto/viam/app/v1/app.proto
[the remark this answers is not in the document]
ehhong: no reason, just added!

michaellee1019 (approved):

PR #405, opened 2023-11-28
add BuildService

abe-winter:
## What changed
- Add BuildService with RPCs for starting and managing builds
## Reviewer notes
- We've marked this unstable until the implementation lands


proto/viam/app/build/v1/build.proto
```diff
@@ -0,0 +1,82 @@
...
+  // optional git ref; defaults to 'main'
+  optional string ref = 2;
+  // list of platforms to build
+  repeated string platform = 3;
+  // module_id as prefix:name pair
+  string module_id = 4;
+}
+
+message StartBuildResponse {
+  string build_id = 1;
```
michaellee1019, line 34: Shouldn't this be `repeated string build_id`? Based on the protos below, it feels like there will be a `build_id` generated per `platform`.

Or is JobInfo supposed to have `repeated string platform`

proto/viam/app/build/v1/build.proto
```diff
@@ -0,0 +1,82 @@
...
+  rpc ListJobs(ListJobsRequest) returns (ListJobsResponse);
+}
+
+message StartBuildRequest {
+  // repo to build
+  string repo = 1;
+  // optional git ref; defaults to 'main'
+  optional string ref = 2;
+  // list of platforms to build
+  repeated string platform = 3;
```
michaellee1019, line 28:
```suggestion
  repeated string platforms = 3;
```

proto/viam/app/build/v1/build.proto
```diff
@@ -0,0 +1,82 @@
...
+
+message GetLogsRequest {
+  string build_id = 1;
+  string platform = 2;
+}
+
+// GetLogsResponse is a streaming endpoint that may have multiple messages that belong
+// to the same build_step if there are too many bytes to fit into a single gRPC
+// response.
+message GetLogsResponse {
```
michaellee1019, line 45: This might need to be a streaming API. How long are the logs? I forget our max length for a GRPC request, but given that we don't know the length and could get really long... Its probably best to just make it streaming now and avoid issues later.

proto/viam/app/build/v1/build.proto
```diff
@@ -0,0 +1,82 @@
...
+
+message ListJobsRequest {
+  // module_id as prefix:name pair
+  string module_id = 1;
+  // don't return more than max_jobs_length jobs
+  // if not present, return all jobs.
+  optional int32 max_jobs_length = 2;
+  // only return jobs that match this build id
+  // if not present, return all jobs.
+  optional string build_id = 3;
```
michaellee1019, line 76: Whats the equivalent CLI command for this? I assume you'd have an optional `--build_id` param.

In our apis, we usually have separate Get and List apis. ListAPIs can have filters but usually not including filter using the primary key to get a single item. Usually have build_id as the only parameter to `GetJobRequest`. Any reason to stick with a single api?

michaellee1019 (changes requested):
Nice first pass! Some general suggestions / questions here. Let me know if I'm off base because I haven't seen the whole system in code yet.

proto/viam/app/build/v1/build.proto
abe-winter: the model is 'one build contains multiple jobs'. each job represents one platform.

(and this maps idiomatically onto most of the vendors we've shopped)

proto/viam/app/build/v1/build.proto
abe-winter: getlogs api returns a stream

proto/viam/app/build/v1/build.proto
zaporter-work: Fixed. Good idea.

proto/viam/app/build/v1/build.proto
zaporter-work: The current impl is that each "build" kicks off multiple jobs (one per platform)

proto/viam/app/build/v1/build.proto
zaporter-work: It is a streaming API, that is what the comment says. Did I misunderstand your comment?

proto/viam/app/build/v1/build.proto
abe-winter: we're planning to flesh out CLI later

but in first release sticking to things that are strictly necessary to do this in CI

can be budged from that if your gut says we need a get

zaporter-work: This relates to the single buildid -> multiple platforms pattern (which might not be intuitive...)
This will return info about just the jobs that were kicked off for that build.
Ex: `viam module build list --id xyz123` (will just return info for the 3-4 architectures the module was built on)

proto/viam/app/build/v1/build.proto
michaellee1019: Makes sense now. And no plans for a job_id. Its the `build_id + platform` that makes a job unique?

proto/viam/app/build/v1/build.proto
zaporter-work: Good question. The reason I wrote it like this is primarily laziness (the build step will be ~10 bytes and the data string will be ~4MB so repeating the build step  a few times didn't really seem like a big deal). However, if this api is also confusing, that is bad. I will fix this

zaporter-work: > Does it work without oneof?

Yes, it just wastes a few bytes per message that shares the same build step

michaellee1019: Ok, I learned it isn't required today. I see other APIs that don't do it now too. I feel like the "convention" should be to use `oneof` but I think there are more usages without it than with. So we can pass on this.

michaellee1019:
LGTM! Let me know if you want an approval now, or later (depends on when you want the protos to build).

michaellee1019 (approved):

michaellee1019 (approved):

PR #410, opened 2023-12-07
[APP-2998] Add OrganizationIdentity message in AppService

maxhorowitz:
Goal is to make all secondary orgs on a shared location visible, but to only reveal the org's name if the current user is a member of that org (and just render org ID otherwise).

In order to reveal all orgs that share a location, we need to send the list of all secondary orgs from the AppService server to the client, but we want to filter out information that the current user should not be able to see. We can use the existing `Organization` proto message type, but that includes CMRId (salesforce ID), stripe email, and org name. So if we were to use it, we would need to overwrite those values as empty before sending from server to client. The concern is that this will cause future bugs (because anyone working on the frontend is likely to assume that those values are populated and correct). So, creating a new `OrganizationIdentity` proto message type is more fitting to create in the backend and send to the client.


proto/viam/app/v1/app.proto
```diff
@@ -378,6 +378,15 @@ message DeleteOrganizationMemberResponse {}
 // Location
 //
 
+// Used for rendering an organization's information on the frontend (limited
+// to id, name, or both).
+message OrganizationIdentity {
+  // Organization ID.
+  string organization_id = 1;
+  // Organization name.
+  string organization_name = 2;
```
jr22, line 387: making these `organization_id` and `organization_name` are somewhat redundant. could could probably make it `id` and `name` instead

jr22 (approved):
one optional comment but otherwise LGTM!j i think youll have to revert the manually compiled protos and regenerate them through the workflow (i think??)

PR #411, opened 2023-12-07
RSDK-4789 - Rename MoveOnGlobeNew to MoveOnGlobe

nicksanford:


proto/viam/service/motion/v1/motion.proto
```diff
@@ -33,20 +33,16 @@ service MotionService {
     };
   }
 
+  // Generate and begin executing an execution to move a component
+  // to a specific GPS coordinate.
+  // May replan to avoid obstacles & account for location drift.
+  // Creates a new plan upon replanning.
```
raybjork, line 39: I think this is a little confusing without a greater description of what these terms mean, which IMO is what our docs page is for.  Maybe we should just keep it general rather than implying things about what happens under the hood?

raybjork: Actually none of the other methods even have comments on them so I think its maybe fine to not add this?  I would hope that people would refer to our docs page for usage, etc.

raybjork (changes requested):
Just a comment on the comment

proto/viam/service/motion/v1/motion.proto
nicksanford: discussed offline, I believe these comments are used by some SDKs

raybjork (approved):
I still think you should remove the comment.  I just made RSDK-6055 to track this and think we should just do everything in this file as a part of that ticket.  Either way though you have the go ahead to merge

PR #416, opened 2023-12-15
[APP-3434] MRR Deletes: Add API protos

hexbabe:
https://viam.atlassian.net/browse/APP-3434

Let me know if I missed anything! This is my first API/proto change


proto/viam/app/v1/app.proto
```diff
@@ -951,6 +953,15 @@ message ListRegistryItemsResponse {
   repeated RegistryItem items = 1;
 }
 
+message DeleteRegistryItemRequest {
+  // The id of the item (formatted as prefix:name where prefix is the owner's orgid or namespace)
```
hexbabe, line 957: Is this true? I mirrored `module_id`

michaellee1019: Yup! that is correct.

michaellee1019 (approved):

proto/viam/app/v1/app.proto
ehhong: technically package type is unnecessary because orgID <> name are unique. but i'm fine leaving this in just for protection. i imagine the get registry item call will use the same request shape?

michaellee1019: Great point. I was out of sync with that decision and this tech spec. I think lets remove type then, less to cleanup later. I can re-approve to regenerate the code after the proto is updated.

hexbabe:
How do we run the final `Check Merge Requirements` check?

hexbabe:
Would Michael un-approving then re-approving the PR not retrigger the compilation?

ehhong:
> > Emily: you made proto changes after protos were compiled though
> > Michael: I can re-approve to regenerate the code after the proto is updated.
> 
> Would Michael un-approving then re-approving the PR fix this?

it won't if the `proto-compiled` label is still there. i also recommend clearing the generated files from your PR so there aren't merge conflicts when it tries to regenerate them

hexbabe:
Got it, so 

1. I will remove that label
2. I will remove all the generated files and push again
3. Emily or Michael will approve again

hexbabe:
Ok... that was a non-traditional way of doing the above, but I have removed the tag and the generated files

ehhong (approved):
lol. excellent work! looks good to me :)

michaellee1019:
LGTM!

hexbabe:
Omg how do we addressing resolving conflicts in generated files?

ehhong:
> Omg how do we addressing resolving conflicts in generated files?

you'll have to rebase main and start the process over again :( sorry it's so finicky and complicated!

hexbabe:
Ok the commit history doesn't say it, but I merged in changes from main during my force commit

michaellee1019 (approved):

hexbabe:
Cool gonna merge. Thanks for troubleshooting this with me!

PR #422, opened 2024-01-02
RSDK-6117: updating nmea fix to int. 

susmitaSanyal:


susmitaSanyal:
solution is not valid. Figuring it out. open anther PR.

PR #423, opened 2024-01-02
Rsdk-6117: getaccuracy v3

susmitaSanyal:
converting gga fix response to int

### DIFF
--- a/proto/viam/component/movementsensor/v1/movementsensor.proto
+++ b/proto/viam/component/movementsensor/v1/movementsensor.proto
@@ -9,6 +9,26 @@ import "google/protobuf/struct.proto";
 option go_package = "go.viam.com/api/component/movementsensor/v1";
 option java_package = "com.viam.component.movementsensor.v1";
 
+// NmeaGGAFix is deprecated
+// NmeaGGAFix includes fix quality of a GPS which is obtained from the GGA message.
+// More information on GGA messages here
+// https://receiverhelp.trimble.com/alloy-gnss/en-us/NMEA-0183messages_GGA.html
+// Generally a fix of 1 or 2 lends to large position errors, ideally we want a
+// fix of 4-5. Other fixes are unsuitable for outdoor navigation.
+// The meaning of each fix value is documented here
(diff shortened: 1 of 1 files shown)


proto/viam/component/movementsensor/v1/movementsensor.proto
```diff
@@ -17,15 +18,15 @@ option java_package = "com.viam.component.movementsensor.v1";
...
-  NMEA_GGA_FIX_SIMULATION = 8;
+  NMEA_GGA_FIX_INVALID_UNSPECIFIED = 0 [deprecated = true];
+  NMEA_GGA_FIX_GNSS = 1 [deprecated = true];
+  NMEA_GGA_FIX_DGPS = 2 [deprecated = true];
+  NMEA_GGA_FIX_PPS = 3 [deprecated = true];
+  NMEA_GGA_FIX_RTK_FIXED = 4 [deprecated = true];
+  NMEA_GGA_FIX_RTK_FLOAT = 5 [deprecated = true];
+  NMEA_GGA_FIX_DEAD_RECKONING = 6 [deprecated = true];
+  NMEA_GGA_FIX_MANUAL = 7 [deprecated = true];
+  NMEA_GGA_FIX_SIMULATION = 8 [deprecated = true];
```
randhid, line 29: So we immediately need to have another pr to remove it?

susmitaSanyal: maybe. I'm not sure. I tried removing it, buf doesn't like it. seems like deprecating it is the only way to not get error from buf.

randhid: if it's a breaking change you need Simone or Steve past the errors.

Let's do te following - let's remove the enum for now, ask for SImone to review, and then add the integer in a new pr.

Next time - use a local version of the API in your rdk and the sdks to code, build and tests and then merge things in in quick succession to each other.

susmitaSanyal:
@Simberific, this needs to be approved by you since this is an api breaking change.

Simberific (approved):

PR #425, opened 2024-01-04
APP-3557: Add Registry Item External Usage, Updated, Created fields

michaellee1019:
Adds fields that return external usage of the registry item (both orgs and robots) as well as Created and Updated fields.


proto/viam/app/v1/app.proto
```diff
@@ -912,6 +912,15 @@ message RegistryItem {
     ModuleMetadata module_metadata = 11;
     MLModelMetadata ml_model_metadata = 12;
   }
+
+  // The total number of robots using this item outside of the owning org
+  int64 total_external_robot_usage = 13;
+  // The total number of organizations using this item outside of the owning org
+  int64 total_external_organization_usage = 14;
```
ehhong, line 919: is `total_external_robot_usage` = `total_robot_usage` - <# of robots in your org using the item>?
similarly `total_external_organization_usage` = `total_organization_usage` - 1?

proto/viam/app/v1/app.proto
```diff
@@ -912,6 +912,15 @@ message RegistryItem {
     ModuleMetadata module_metadata = 11;
     MLModelMetadata ml_model_metadata = 12;
   }
```
ehhong, line 914: we will likely different types of metadata as we support more types in the registry, it's a breaking change either way but how would we do this if `metadata` is in the middle of the struct? not sure what the pb best practices are here

michaellee1019: Its totally acceptable to have protobufs where their fields are out of order when numbered. In the future it will probably look like this, and will work:

  oneof metadata {
    ModuleMetadata module_metadata = 11;
    MLModelMetadata ml_model_metadata = 12;
    SlamMapMetadata slam_metadata = 17;
  }

Remember that these fields are just use in serialize/deserialize of the message. In transit the payload is sent like `1=value,2=value` which allows the contents of the message to be shorter by not using field names. Also field names can change without breaking.

Anyways, I think `oneof` will be fine. The message with get tranferred with either 11, 12, or 17. and the gRPC client code will decipher that and make sure one is returned.

Also not sure we have other options than make the additional fields 50,51,52,53. Because we don't know how many types will be added over the years.

hexbabe: Don't we preserve the serialization numbers (i.e. keeping metadata in the middle of the struct) in order to ensure backwards compatibility?

proto/viam/app/v1/app.proto
michaellee1019: Yup that is correct. There aren't fields for `total_internal_robot/organization_usage` so you can subtract external from total to get internal

hexbabe (approved):

proto/viam/app/v1/app.proto
michaellee1019: > Not 100% sure: Don't we preserve the serialization numbers (i.e. keeping metadata in the middle of the struct) in order to ensure backwards compatibility?

Yeah we should for sure not renumber the metadata numbers as that will break clients.

ehhong: ah didn't realize you could do this! this sounds good to me then. an optional nit would be to put the new usage fields directly under the previous usage fields just so they're all in one place
```
...
int64 total_robot_usage = 9;
// The total number of organizations using this item
int64 total_organization_usage = 10;
// The total number of robots using this item outside of the owning org
int64 total_external_robot_usage = 13;
// The total number of organizations using this item outside of the owning org
int64 total_external_organization_usage = 14;

// Type-specific metadata
oneof metadata {
  ModuleMetadata module_metadata = 11;
  MLModelMetadata ml_model_metadata = 12;
}
```

michaellee1019: SGTM! will do that.

michaellee1019:
Hey @hexbabe / @ehhong can one of you re-approve the PR? The compile protos job didn't kick off.

ehhong (approved):
gotchu

PR #429, opened 2024-01-17
RSDK-6185 Move LogEntry to common

benjirewis:
[RSDK-6185](https://viam.atlassian.net/browse/RSDK-6185)

Per a post hoc scope edit, we should move `LogEntry` to common now that it's used by both `app` and `rdk`.

More context:
> It makes more sense semantically, and doing so avoids `robot.proto` importing `app.proto` which imports some odd proto files like `tagger.proto`. `tagger.proto`, for example, requires extra Golang dependencies to build properly (namely [protoc-gen-gotag](http://github.com/srikrsna/protoc-gen-gotag)). This actually shouldn't be backward breaking as the structure of the API is identical despite a new import path; see [when we moved](https://docs.google.com/document/d/1whBUsuclatpUi7plzVLl8E0xYXvSH9mB0AlD9SI0la0/edit#heading=h.4twklylkxwe0) `DoCommandRequest/Response` to `common.proto`.

[RSDK-6185]: https://viam.atlassian.net/browse/RSDK-6185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dgottlieb (approved):
My understanding is that the lint/compatibility errors here are false positives. I double checked the numbering on the fields.

But probably best to let @cheukt double check.

cheukt (approved):
should be fine, can test by seeing if a rdk robot with this set of changes can upload logs to the cloud

benjirewis:
> can test by seeing if a rdk robot with this set of changes can upload logs to the cloud

Great idea; it works, so seems like the theory of "maintaining shape of API wrt serialization means changes aren't backward breaking". Let me check in with fleet before I merge this, though.

benjirewis:
Going to close this PR. We're going [to revert](https://github.com/viamrobotics/api/pull/431) the original `ModuleLog` changes (as they slightly break Golang SDK clients), and I'll move `LogEntry` to common at a future point in time when there are fewer breaking `api` changes in the mix (deprecation of sensors service + removal of `GetLatestMapInfo` from SLAM).

PR #430, opened 2024-01-18
RSDK-6378 Add proto file for generic service

benjirewis:
[RSDK-6738](https://viam.atlassian.net/browse/RSDK-6378)

Adds a new `proto/viam/service/generic/v1/generic.proto` file.


proto/viam/component/generic/v1/generic.proto
```diff
@@ -8,7 +8,8 @@ import "google/api/annotations.proto";
 option go_package = "go.viam.com/api/component/generic/v1";
 option java_package = "com.viam.component.generic.v1";
 
-// GenericService services all generic commands associated with a robot
+// GenericService services all generic commands for components associated with
```
cheukt, line 11:
```suggestion
// GenericService services all generic components associated with
```

proto/viam/service/generic/v1/generic.proto
```diff
@@ -0,0 +1,20 @@
...
+
+package viam.service.generic.v1;
+
+import "common/v1/common.proto";
+import "google/api/annotations.proto";
+
+option go_package = "go.viam.com/api/service/generic/v1";
+option java_package = "com.viam.service.generic.v1";
+
+// GenericService services all generic commands for services associated with
```
cheukt, line 11:
```suggestion
// GenericService services all generic services associated with
```

proto/viam/service/generic/v1/generic.proto
```diff
@@ -0,0 +1,20 @@
...
+// a robot
+service GenericService {
+  // DoCommand sends/recieves arbitrary commands
```
cheukt, line 14: receives?

proto/viam/component/generic/v1/generic.proto
```diff
@@ -8,7 +8,8 @@ import "google/api/annotations.proto";
...
+// a robot
 service GenericService {
   // DoCommand sends/recieves arbitrary commands
```
cheukt, line 14: receives?

cheukt (changes requested):
small doc updates requested

proto/viam/service/generic/v1/generic.proto
cheukt: word is spelled wrong

proto/viam/component/generic/v1/generic.proto
benjirewis: Ah good idea; didn't commit suggestion because formatting changed a bit.

proto/viam/service/generic/v1/generic.proto
benjirewis: Done.

proto/viam/service/generic/v1/generic.proto
benjirewis: 😮‍💨 gotcha.

cheukt (approved):

proto/viam/service/generic/v1/generic.proto
```diff
@@ -0,0 +1,19 @@
...
+package viam.service.generic.v1;
+
+import "common/v1/common.proto";
+import "google/api/annotations.proto";
+
+option go_package = "go.viam.com/api/service/generic/v1";
+option java_package = "com.viam.service.generic.v1";
+
+// GenericService services all generic services associated with a robot
+service GenericService {
```
maximpertsov, line 12: [q] these just seem identical... have we considered having a "GenericService" in our common namespace that corresponds to both services and components?

benjirewis: Yeah it's a good question. I _think_ unifying to a single `service GenericService` at the proto level would necessitate one client-side and one server-side wrapper for `DoCommand` for both services and components (I'm unaware of an ability to define different senders/receivers for a single proto `service` in a single language). In a world where we've moved past the distinction between services and components in all our SDKs, I think a single proto `service GenericService` would be the correct move. So, in short, I'm not really sure if it's possible to put anything beyond `DoCommandRe*`in our common namespace right now. All ears if I'm misunderstanding the limitations of proto/our current architecture, though!

maximpertsov (approved):

proto/viam/service/generic/v1/generic.proto
benjirewis: Spoke a bit with Maxim offline; it is probably possible to use a single proto service and maybe we could if we were sure services and components were converging in functionality, but the generic component service does have an extra method `GetGeometries`, so they're not even really identical right now. Gonna stick with this design.

PR #414, opened 2023-12-11
RSDK-6185 Add `ModuleLog` method to `RobotService`

benjirewis:
Adds a simple `ModuleLog` method and `ModuleLogRequest/Response`.

PR #431, opened 2024-01-22
Revert "RSDK-6185 Add `ModuleLog` method to `RobotService` (#414)"

benjirewis:
This reverts the proto changes of commit 7f530337450d2e29a0278f7b64bebb8caca30671. We should revert this commit because it slightly breaks gRPC reflection for golang clients (you'll see `TestClientResources` fail to grab the `RobotService` service description from a remote if you upgrade the `rdk` to use a version of `api` that has this commit).


cheukt (approved):

randhid (approved):

randhid:
@benjirewis I think you might need steve's approval unless cheuk has the special power for this repo.

PR #433, opened 2024-01-23
APP-3631: MLModelMetadata version strings to structs

michaellee1019:
Updated protos to changes ML Model version string array to a struct array. This makes the proto consistent with modules and allows additional fields to be safely added over time.


ehhong (approved):

michaellee1019:
I'm going to not do this right now. Instead lets do it when we need additional metadata to be exposed. Whatever project requires it.

PR #436, opened 2024-01-25
[APP-3742] Draft proto for provisioning

ale7714:


proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+
+option go_package = "go.viam.com/api/provisioning/v1";
+
+// ProvisionService is used by viam-agent to retrieve device and robot info.
+service ProvisionService {
+  // GetDeviceStatus is for retrieving the status of the device including networking.
+  rpc GetDeviceStatus(GetDeviceStatusRequest) returns (GetDeviceStatusResponse);
+
+  // SetWifiCredentials is to set the wifi credentials.
+  rpc SetWifiCredentials(SetWifiCredentialsRequest) returns (SetWifiCredentialsResponse);
```
ale7714, line 13: I'm going back and forth about having an AgentNetworking service and moving the wifi related functions there since networking management will be available outside of provisioning in the future. WDYT?

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+  repeated ConnectionInfo wifis = 1;
+}
+
+message ProviosionInfo { 
+  string fragment_id  = 1;
+  string model = 2;
+  string manufacturer = 3;
+}
+
+message ProvisionStatus {
```
ale7714, line 59: I'm not using this yet... just an idea and i don't like the name. If we wanted to add context to device status about the subsystems, maybe we could do something like this.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+message ConnectionAttempt {
+  ConnectionInfo connection_info = 1;
+  string command = 2;
+  string error_msg = 3;
+}
+
+message ConnectionInfo {
+  string type = 1;
+  string sidd = 2;
+  string security = 3;
```
ale7714, line 78: Though this could be useful to highlight in the mobile app insecure wifis

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+  rpc GetDeviceStatus(GetDeviceStatusRequest) returns (GetDeviceStatusResponse);
+
+  // SetWifiCredentials is to set the wifi credentials.
+  rpc SetWifiCredentials(SetWifiCredentialsRequest) returns (SetWifiCredentialsResponse);
+
+  // SetRobotCredentials is to set the robot credentials.
+  rpc SetRobotCredentials(SetRobotCredentialsRequest) returns (SetRobotCredentialsResponse); 
+
+  // GetVisibleWifiNetworks is to retrieve the list of wifi networks that are visible to the device.
+  rpc GetVisibleWifiNetworks(GetVisibleWifiNetworksRequest) returns (GetVisibleWifiNetworksResponse);
```
ale7714, line 19: Should we call this just GetVisibleNetworks?

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
+syntax = "proto3";
+
+package viam.provisioning.v1;
+
+option go_package = "go.viam.com/api/provisioning/v1";
+
+// ProvisionService is used by viam-agent to retrieve device and robot info.
+service ProvisionService {
```
Otterverse, line 8:
```suggestion
service ProvisioningService {
```
To match the name of the package.

proto/viam/provisioning/provisioning.proto
Otterverse: I think they can use the same service. It just won't use the "SetRobotCredentials" call.

proto/viam/provisioning/provisioning.proto
Otterverse: GetNetworkList?

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+
+message SetWifiCredentialsResponse {
+  ConnectionAttempt connection_attempt = 1; 
+}
+
+message SetRobotCredentialsRequest {
+}
+
+message SetRobotCredentialsResponse {
+  CloudConfig cloud = 1;
```
Otterverse, line 43: This has to go in the request, not the response.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+// ProvisionService is used by viam-agent to retrieve device and robot info.
+service ProvisionService {
+  // GetDeviceStatus is for retrieving the status of the device including networking.
+  rpc GetDeviceStatus(GetDeviceStatusRequest) returns (GetDeviceStatusResponse);
+
+  // SetWifiCredentials is to set the wifi credentials.
+  rpc SetWifiCredentials(SetWifiCredentialsRequest) returns (SetWifiCredentialsResponse);
+
+  // SetRobotCredentials is to set the robot credentials.
+  rpc SetRobotCredentials(SetRobotCredentialsRequest) returns (SetRobotCredentialsResponse);
```
Otterverse, line 16:
```suggestion
  rpc SetCloudCredentials(SetCloudCredentialsRequest) returns (SetCloudCredentialsResponse); 
```
We should get ahead of the curve. Smart machines, not robots. :-)

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+
+message SetRobotCredentialsResponse {
+  CloudConfig cloud = 1;
+}
+
+message GetVisibleWifiNetworksRequest {
+}
+
+message GetVisibleWifiNetworksResponse {
+  repeated NetworkInfo wifis = 1;
```
Otterverse, line 50:
```suggestion
  repeated NetworkInfo networks = 1;
```
We may end up putting wired and cellular connections in here eventually.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+}
+
+message ProviosionInfo {
```
Otterverse, line 53:
```suggestion
message ProvisioningInfo { 
```

proto/viam/provisioning/provisioning.proto
Otterverse: Let's not add it until we need it.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+  ConnectionAttempt latest_connection_attempt = 2;
+}
+
+message SetWifiCredentialsRequest {
+  string ssid = 1;
+  string psk = 2;
+}
+
+message SetWifiCredentialsResponse {
+  ConnectionAttempt connection_attempt = 1;
```
Otterverse, line 36: We likely can't send this here, as we won't have tried connecting yet.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+  // GetVisibleWifiNetworks is to retrieve the list of wifi networks that are visible to the device.
+  rpc GetVisibleWifiNetworks(GetVisibleWifiNetworksRequest) returns (GetVisibleWifiNetworksResponse); 
+}
+
+message GetDeviceStatusRequest {
+}
+
+message GetDeviceStatusResponse {
+  ProviosionInfo provision_info = 1;
+  ConnectionAttempt latest_connection_attempt = 2;
```
Otterverse, line 27: Also need some more general status stuff here. A device might be provisioned (have device creds) but not online, or vice versa. Likewise, there may be other errors to report.

Suggestions:

bool has_device_credentials
bool is_online
repeated string errors

Not sure if ConnectionAttempt should be its own separate message or not. I think just "NetworkInfo" directly, after my comments on modifying that to include the error directly.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+message SubsystemStatus {
+  string name = 1;
+  string status = 2;
+  string error_msg = 3;
+}
+
+message ConnectionAttempt {
+  NetworkInfo network_info = 1;
+  string command = 2;
+  string error_msg = 3;
```
Otterverse, line 72: I don't know what "command" would be here. I think just put error_msg (or "last_error") inside NetworkInfo and use it directly. Then errors remain associated with a specific network that way, in case people need to try multiple wifis.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+}
+
+message NetworkInfo {
+  string type = 1;
+  string sidd = 2;
```
Otterverse, line 77:
```suggestion
  string ssid = 2;
```

proto/viam/provisioning/provisioning.proto
Otterverse: Good idea!

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,90 @@
...
+  string security = 3;
+  int32 signal = 4;
+}
```
Otterverse, line 80: Would add one more:

`bool connected = 5`

Otterverse (changes requested):
Thanks for the head start on this. Left lots of small comments/changes for now. There will probably be more/other changes as we try to implement this. Looking at the size of the PR now, it's pretty significant, so I don't think it's worth trying to merge it into the other API PR like we mentioned yesterday.

PR #438, opened 2024-01-26
[APP-3742] Provisioning proto

ale7714:
Followup from https://github.com/viamrobotics/api/pull/436 which got closed after I deleted the branch with updates on the viam agent.


proto/viam/provisioning/provisioning.proto
Otterverse, line 1: This filename/path needs to match the versioned layout scheme along with the package and go_package settings below. It needs to be in a v1 folder `proto/viam/provisioning/v1/provisioning.proto`

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
+syntax = "proto3";
+
+package viam.provisioning.v1;
+
+option go_package = "go.viam.com/api/provisioning/v1";
+
+// ProvisioningService is used by viam-agent to retrieve device and smart machine info.
+service ProvisioningService {
+  // GetDeviceStatus is for retrieving the status of the device including networking.
+  rpc GetDeviceStatus(GetDeviceStatusRequest) returns (GetDeviceStatusResponse);
```
Otterverse, line 10: We should decide if this is a "Device" or a "SmartMachine" and use the same term in the RPCs here.

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+
+message GetDeviceStatusResponse {
+  ProviosioningInfo provisioning_info = 1;
+  bool has_device_credentials = 2;
+  bool is_online = 3;
+  NetworkInfo latest_connection_attempt = 4;
+  repeated string errors = 5;
+}
+
+message SetWifiCredentialsRequest {
```
Otterverse, line 33: As we called it "GetNetworkList()" for the other RPC, this maybe should be "SetNetworkCredentials" as well to match. In which case it should have a "type" field as #1 I think. Thoughts?

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+}
+
+message GetNetworkListRequest {
+}
+
+message GetNetworkListResponse {
+  repeated NetworkInfo networks = 1;
+}
+
+message ProviosioningInfo {
```
Otterverse:
```suggestion
message ProvisioningInfo { 
```

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+
+message ProviosioningInfo { 
+  string fragment_id  = 1;
+  string model = 2;
+  string manufacturer = 3;
+}
+
+message NetworkInfo {
+  string type = 1;
+  string sidd = 2;
```
Otterverse, line 63:
```suggestion
  string ssid = 2;
```

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+
+  // GetNetworkList is to retrieve the list of networks that are visible to the device.
+  rpc GetNetworkList(GetNetworkListRequest) returns (GetNetworkListResponse); 
+}
+
+message GetDeviceStatusRequest {
+}
+
+message GetDeviceStatusResponse {
+  ProviosioningInfo provisioning_info = 1;
```
Otterverse:
```suggestion
  ProvisioningInfo provisioning_info = 1;
```

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+
+option go_package = "go.viam.com/api/provisioning/v1";
+
+// ProvisioningService is used by viam-agent to retrieve device and smart machine info.
+service ProvisioningService {
+  // GetDeviceStatus is for retrieving the status of the device including networking.
+  rpc GetDeviceStatus(GetDeviceStatusRequest) returns (GetDeviceStatusResponse);
+
+  // SetWifiCredentials is to set the wifi credentials.
+  rpc SetWifiCredentials(SetWifiCredentialsRequest) returns (SetWifiCredentialsResponse);
```
Otterverse, line 13: Should this be `SetNetworkCredentials` to match the terminology of `GetNetworkList`?

proto/viam/provisioning/provisioning.proto
```diff
@@ -0,0 +1,78 @@
...
+  // GetNetworkList is to retrieve the list of networks that are visible to the device.
+  rpc GetNetworkList(GetNetworkListRequest) returns (GetNetworkListResponse); 
+}
+
+message GetDeviceStatusRequest {
+}
+
+message GetDeviceStatusResponse {
+  ProviosioningInfo provisioning_info = 1;
+  bool has_device_credentials = 2;
```
Otterverse, line 27: This either needs to be "smart_machine_credentials" OR we should change the other RPC to be "SetDeviceCredentials" I'm not sure which directly is better. I like the shortness of "device" but it's also very generic.

Otterverse (changes requested):
As noted, the filename needs updating. Also though, after you build (and lint) locally, see if the linter changed the .proto file itself. Currently it does on my system.

proto/viam/provisioning/provisioning.proto
ale7714: Agree with you, End up going for smart_machine, because device is too generic I think. The ideal short term would be robot 🙈 . I think smart_machine correctly conveys the intent even if it's long.

Otterverse (commented):
This looks good to me at this point, but let's wait to merge it until after we actually write some code to use it, in case more needs to be added.

PR #442, opened 2024-01-31
APP-3385: Logs tab: Add Filtering options to sidebar

Oabraham1:
[Ticket](https://viam.atlassian.net/browse/APP-3385)

In this PR, I added an extra field to the `GetRobotPartLogsRequest` message called `levels`. The goal of `levels` is to add an extra filtering option based on the level of a particular log in the database when fetching the logs.

This PR has to be merged before [this pr](https://github.com/viamrobotics/app/pull/3530) can be closed out since the `request.level` filed needs to be provided in the `app/service/server.go` file.


Oabraham1:
I was also thinking I could just make the `filter` field an array of strings and use that for the filtering @michaellee1019

michaellee1019:
The changes LGTM. Make sure to implement in a backwards compatible way on the backend: If an empty list is passed in then we should default to returning all levels.

I'm checking to see if we need to scope this.

stuqdog (approved):
Looks reasonable to me, agree with @njooma re: SDK perspective

proto/viam/app/v1/app.proto
```diff
@@ -603,10 +603,20 @@ message GetRobotPartResponse {
 }
 
 message GetRobotPartLogsRequest {
+  enum LogLevel {
```
RoxyFarhad, line 606: can you move this out of the message for LogsRequest? It may be useful for other requests

RoxyFarhad (approved):
approving under assumption that you move the enum out of the message

proto/viam/app/v1/app.proto
Oabraham1: Done

DTCurrie:
> The changes LGTM. Make sure to implement in a backwards compatible way on the backend: If an empty list is passed in then we should default to returning all levels.
> 
> I'm checking to see if we need to scope this.

Can we document this is the expected behavior in some comments?

Oabraham1:
> > The changes LGTM. Make sure to implement in a backwards compatible way on the backend: If an empty list is passed in then we should default to returning all levels.
> > I'm checking to see if we need to scope this.
> 
> Can we document this is the expected behavior in some comments?

I could add it to the ticket description or comments if that's what you mean?

DTCurrie:
> > > The changes LGTM. Make sure to implement in a backwards compatible way on the backend: If an empty list is passed in then we should default to returning all levels.
> > > I'm checking to see if we need to scope this.
> > 
> > 
> > Can we document this is the expected behavior in some comments?
> 
> I could add it to the ticket description or comments if that's what you mean?

Actually, this change is going to require a scope doc. We will have to let the Docs team know and clarify exactly how we expect this API to behave. We expect this API to be used externally, so we have to clearly communicate any changes we are making.

Oabraham1:
[Change scope doc](https://docs.google.com/document/d/1YLAixH7DS7PvNZUfsxx9DvqqAgH6W80_7xtGVQLg9aM/edit?usp=sharing) was approved

proto/viam/app/v1/app.proto
```diff
@@ -604,9 +604,10 @@ message GetRobotPartResponse {
 
 message GetRobotPartLogsRequest {
   string id = 1;
-  bool errors_only = 2;
+  bool errors_only = 2 [deprecated = true];
```
DTCurrie, line 607: Should we include a comment pointing to the ticket to remove this down the road?

DTCurrie (approved):
✅

proto/viam/app/v1/app.proto
Oabraham1: Added the comment

PR #445, opened 2024-02-06
APP-3197 Update GetRobotAPIKeys to return authorizations

ehhong:
This change supports the connect tab API keys section, which is being re-designed to display the authorizations on an API key ([slack convo](https://viaminc.slack.com/archives/C05V88RQUJH/p1707247996433699)).


proto/viam/app/v1/app.proto
```diff
@@ -669,6 +669,7 @@ message APIKey {
 
 message GetRobotAPIKeysResponse {
   repeated APIKey api_keys = 1;
+  repeated APIKeyWithAuthorizations api_keys_with_authorizations = 2;
```
ehhong, line 672: should I just fully break the API and do this?
```proto
message GetRobotAPIKeysResponse {
  repeated APIKeyWithAuthorizations api_keys = 1;
}
```

jr22: i think fully break API

ehhong: updated

jr22 (approved):
:shipit: 

i think steve or someone with equivalent perms will have to merge this for you

PR #448, opened 2024-02-07
APP-3385: Regenerate Protos

Oabraham1:
We overwrote changes that were resolved from merge conflicts which meant that the client generated code does not work in app. This is just an empty commit to regenerate client code as it should be on main.


ethanlook (approved):
Approving so we can see what the generated changes are. Let's wait to merge until we validate that this works in app.

Oabraham1:
> Approving so we can see what the generated changes are. Let's wait to merge until we validate that this works in app.

We just validated this locally

PR #449, opened 2024-02-07
APP-3385: Updated comment describing levels field

Oabraham1:
We need to regenerate protos because the buf-push action fails when there are no changes on main.


Oabraham1:
I needed to rebase onto main, so I did that locally and force pushed _before_ protos had been generated.

ethanlook (approved):

PR #454, opened 2024-02-20
RSDK-6661 Add service rpc

maximpertsov:
Forgot to add this as part of https://github.com/viamrobotics/api/pull/453

Blocks: https://github.com/viamrobotics/rdk/pull/3599


proto/viam/robot/v1/robot.proto
```diff
@@ -120,6 +120,13 @@ service RobotService {
       post: "/viam/api/v1/log"
     };
   }
+
+  // GetCloudMetadata returns app-related information about the robot.
+  rpc GetCloudMetadata(GetCloudMetadataRequest) returns (GetCloudMetadataResponse) {
+    option (google.api.http) = {
+      get: "/viam/api/v1/cloud_metadata"
+    };
+  }
```
maximpertsov, line 129: would we be open to dropping the `Get` prefix and just making this rpc and it's related request and response just `CloudMetadata`?

cheukt: no, let's not - unless we do it for every other proto, we should stay consistent with the rest of the repo

cheukt (approved):

PR #455, opened 2024-03-07
[APP-4187] API change to send failed fragment overwrite logs to robot

piokasar:


proto/viam/app/v1/robot.proto
```diff
@@ -37,6 +37,7 @@ message RobotConfig {
   repeated ModuleConfig modules = 9;
   optional bool disable_partial_start = 10;
   repeated PackageConfig packages = 11;
+  repeated string overwrite_fragment_errors = 12;
```
michaellee1019, line 40:
```suggestion
  repeated AppValidationStatus overwrite_fragment_status = 12;
```
That way additional fields could be added in the future if needed. Also aligns to use the same concept across the config.

michaellee1019 (changes requested):

michaellee1019 (approved):

PR #460, opened 2024-03-13
APP-2985: Protos for endpoint to register fusionauth applications

jr22:
APP-2985

Created protos to register fusion auth apps. Not sure if `end_user` is the right place for this? was thinking maybe but still getting acquainted with this stuff so open to feedback if it should be moved elsewhere.


app/v1/end_user.pb.go
```diff
@@ -182,6 +182,123 @@ func (*AcceptLegalResponse) Descriptor() ([]byte, []int) {
...
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
+
+	ApplicationName string   `protobuf:"bytes,1,opt,name=application_name,json=applicationName,proto3" json:"application_name,omitempty"`
+	ApplicationId   string   `protobuf:"bytes,2,opt,name=application_id,json=applicationId,proto3" json:"application_id,omitempty"`
+	OrgId           string   `protobuf:"bytes,3,opt,name=org_id,json=orgId,proto3" json:"org_id,omitempty"`
+	OriginUris      []string `protobuf:"bytes,4,rep,name=origin_uris,json=originUris,proto3" json:"origin_uris,omitempty"`
+	RedirectUris    []string `protobuf:"bytes,5,rep,name=redirect_uris,json=redirectUris,proto3" json:"redirect_uris,omitempty"`
+}
```
jr22, line 195: @ohEmily could you weigh in on these params and if they make sense?

ohEmily: Curious about `omitempty`. I thought we shouldn't do `omitempty` for `OriginUris`, `RedirectUris`, `OrgId` and `ApplicationName`, unless we intend to do app-side validation for those. Wouldn't `omitempty` mean that the fields are optional? Out of these, I believe `ApplicationId` is the only optional one that should have `omitempty`.

proto/viam/app/v1/end_user.proto
```diff
@@ -11,6 +11,9 @@ service EndUserService {
 
   // Marks that the specified user has accepted end user license agreements.
   rpc AcceptLegal(AcceptLegalRequest) returns (AcceptLegalResponse);
+
+  // Allows users to register third party applications using Viam linked to the indicated organization
+  rpc RegisterApplication(RegisterApplicationRequest) returns (RegisterApplicationResponse);
```
ohEmily, line 16: [minor] can we make it `RegisterAuthApplication`?

proto/viam/app/v1/end_user.proto
```diff
@@ -23,3 +26,13 @@ message IsLegalAcceptedResponse {
 message AcceptLegalRequest {}
 
 message AcceptLegalResponse {}
+
+message RegisterApplicationRequest {
+  string application_name = 1;
+  string application_id = 2;
```
ohEmily, line 32: sorry I think maybe my comment was misplaced. I think this one should be marked `optional`, although thinking about it more, maybe we should just omit it entirely and force using an auto-generated application ID. The reason is that this is required to be globally unique, and then we don't have to handle customers using an application ID like "app" and wrapping the FusionAuth response to give a detailed exception.

ohEmily (approved):
LGTM but I think we should remove `application_name` to save ourselves the additional effort. To answer your open question, I think putting this in `end_user.proto` is perfect.

proto/viam/app/v1/end_user.proto
jr22: from the ticket description: 
> Our gRPC endpoint should have application name and redirect URIs should be required parameters. (i.e. don’t use FusionAuth’s built in feature of generating a random client ID)

was thinking this meant we wanted them to pass an application id from the parenthesis, may have misunderstood. can remove if thats not the case!

jr22:
> LGTM but I think we should remove `application_name` to save ourselves the additional effort. To answer your open question, I think putting this in `end_user.proto` is perfect.

assume you mean `application_id` not `application_name`?

ohEmily (approved):
Yes I meant `application_id`. Sorry about that!

piokasar (approved):

piokasar (approved):

PR #462, opened 2024-03-18
APP-1935 Show tabular/data storage costs in invoice

ehhong:


proto/viam/app/v1/billing.proto
```diff
@@ -71,6 +71,8 @@ message GetCurrentMonthUsageResponse {
   double discount_amount = 8;
   double total_usage_with_discount = 9;
   double total_usage_without_discount = 10;
+  double image_data_cloud_storage_usage_cost = 11;
```
RoxyFarhad, line 74: Its unclear to me what this means from reading these fields. Is this tabular/binary separation?

ehhong: yes-ish. we decided to use the language used on our [pricing page](https://www.viam.com/product/pricing). it's not juust tabular because it includes tabular data not used for data/ML. basically general = tabular + packages + history + logs

RoxyFarhad: Aren't we breaking down all the usage costs into their own types? can't we just add a 1:1 mapping for that? Otherwise it just seems unclear?

ehhong: we are on the backend, but the [designs](https://www.figma.com/file/2hs9zGWN9nTLqcm8pR38ip/Management?type=design&node-id=3218-23500&mode=design&t=BWezJpcITxcUUzqh-0) shiw an image/general split. I want to avoid doing calculations on the frontend, but if we think we will just display all categories in the future, I'd consider adding a 1:1 mapping. that's a product q though

ehhong: one thing to consider: we are moving from "cloud storage" --> "data management" in our copy. do we want that change to be reflected in this proto? our backend models are all called "cloud storage"

RoxyFarhad: hm yeah, I think lets leave the BE models for now because the descriptions could change a lot over time.

RoxyFarhad (approved):

proto/viam/app/v1/billing.proto
jr22: i would lean towards doing the summing on the frontend? feels more future proof for if we need any breakdown?

jr22 (commented):
FYI be sure to notify relevant stakeholders about API change

proto/viam/app/v1/billing.proto
ehhong: we would have to sum on the billing page, ops billing dashboard, and invoice, which feels more error prone to me -- @nataliajacobowitz how likely do you think this categorization might change?

jr22: oh i would think that on the ops dashboard, we would want to display the full breakdown (more useful from an ops perspective and we can always sum ourselves)? also, if we have eventually have different costs per different type of cloud storage, i also would think we'd want to display that on the invoice?

michaellee1019:
> FYI be sure to notify relevant stakeholders about API change

I think it actually requires a scope :( All public API changes do. It can be super quick and short one.

proto/viam/app/v1/billing.proto
RoxyFarhad: What about having a structure that encompasses all the "general costs". I do think we want to be returning a single value to make sure all places are showing the same thing. But I agree having a breakdown by type is helpful too and would protect us in the future if we wanted to split these types out more.

ehhong:
> > FYI be sure to notify relevant stakeholders about API change
> 
> I think it actually requires a scope :( All public API changes do. It can be super quick and short one.

@michaellee1019 Julia asked Steve what we should do in this situation, Steve said sending a quick email to stakeholders (Eliot, Natalia, Emily, Devin) was sufficient. I just sent an email out about this change

michaellee1019:
> @michaellee1019 Julia asked Steve what we should do in this situation, Steve said sending a quick email to stakeholders (Eliot, Natalia, Emily, Devin) was sufficient. I just sent an email out about this change

Excellent, sounds good!

proto/viam/app/v1/billing.proto
ehhong: just thought of this last night -- I don't think we publicize anywhere what exactly constitutes general "cloud storage" or "data management" costs. adding the breakdown here would also publish that information.

 > on the ops dashboard, we would want to display the full breakdown
regarding this, if we want more info on the ops dashboard I would push for creating a separate endpoint. ops dashboard is strictly for internal use and I wouldn't want to get that wrapped up in public APIs or trying to keep these views consistent

RoxyFarhad: yeah, then Im not sure how else we could it apart from this way :)

jr22: > i don't really want to include the full breakdown because i think it's early optimizing, we don't even know if the full breakdown is stable

i would say the same thing for the image vs non image. at least by having the breakdown, we are giving ourselves the flexibility to change what goes into which category later. if we _had_ broken down what goes into cloud storage earlier, we wouldnt even have to make this API change in the first place. i dont think i'm gonna be convinced the other way but we can agree to disagree if its 2 to 1.

RoxyFarhad: For early optimization the more granular break down makes sense for sure. That being said, we don't want to show the breakdown to users. I'm not sure I am sure what the middle ground here is but I think this going to be somewhat of a product decision? if the latter is definitely not then I think that we have to have more generalized names?

ehhong: thanks guys for the input. I'm now convinced that the granular categories would be better (if we can publish them), otherwise we'll have to do image/general. I'm tracking down some ops / GTM people right now, will let you know when I have an update!

ehhong: okay, discussed offline: decided on two subcategories:
```proto
double binary_data_cloud_storage_usage_cost = 12;
double other_cloud_storage_usage_cost = 13;
```

PR #467, opened 2024-03-29
RSDK-6829 - Add machine part id to ResourceName message

cheukt:
RDK change: https://github.com/viamrobotics/rdk/pull/3757


proto/viam/common/v1/common.proto
```diff
@@ -15,6 +15,7 @@ message ResourceName {
   string type = 2;
   string subtype = 3;
   string name = 4;
+  optional string machine_part_id = 5;
```
maximpertsov, line 18: why optional?

cheukt: everything else in the resource name should always be filled out, so using optional here to be a bit more clear about the optionality. It doesn't seem like we are super consistent about this though, so could see making it a non-optional field as well

maximpertsov: yeah, that was gonna be my next question - it doesn't seem like we are consistent with using the `optional` keyword vs relying on implicit optionality in protos (e.g. this field will just be a blank string)

maximpertsov: i don't have have a strong opinion on this, so won't block - I'm okay making this explicitly `optional`

maximpertsov (approved):

proto/viam/common/v1/common.proto
cheukt: I'd like to start using optional a bit more, I think it makes it easier to tell what the response shape is like by just reading the proto

maximpertsov (approved):

proto/viam/common/v1/common.proto
benjirewis, line 18: Do we not also want the machine ID? Or just the part ID?

cheukt: just part id for now, that's what was scoped and most likely to be useful - user can get cloud metadata if they need more of that stuff

benjirewis (approved):

PR #482, opened 2024-04-17
Revert machine part id

cheukt:
after getting a sense of the problem, machine part id is likely not the best path forward


ale7714 (approved):

cheukt:
@stevebriskin can you merge?

PR #486, opened 2024-04-17
APP-4415: Add machine cost protos to billing endpoint

jr22:
APP-4415

Adding in params for monthly machines + cost per machine per month to `GetCurrentMonthUsageResponse` so we can show those on the billing page. 

Designs for page: 
https://www.figma.com/file/2hs9zGWN9nTLqcm8pR38ip/Management?type=design&node-id=6622-7383&mode=design


RoxyFarhad (approved):

proto/viam/app/v1/billing.proto
```diff
@@ -71,6 +71,8 @@ message GetCurrentMonthUsageResponse {
   double discount_amount = 8;
   double total_usage_with_discount = 9;
   double total_usage_without_discount = 10;
+  double monthly_machines_usage_cost = 11;
+  double cost_per_machine_per_month = 12;
```
stevebriskin, line 75: This field seems unnecessary for the same reason we don't have it for other resources.

proto/viam/app/v1/billing.proto
```diff
@@ -71,6 +71,8 @@ message GetCurrentMonthUsageResponse {
   double discount_amount = 8;
   double total_usage_with_discount = 9;
   double total_usage_without_discount = 10;
+  double monthly_machines_usage_cost = 11;
```
stevebriskin, line 74: "monthly" is redundant since "Month" is in the message name

nit: "machine_usage_cost" can be vague as it can mean the sum of above costs. something like "total_per_machine_usage_cost" is clearer to me. I don't feel strongly about this.

proto/viam/app/v1/billing.proto
jr22: chatted offline, removing and will omit from FE design as well. users can look at invoice for the sku information like this.

proto/viam/app/v1/billing.proto
jr22: the "total" is a bit confusing because the other fields that start with "total" represent the full total cost whereas this could be one of many SKUs in the future potentially (although i know right now its the only sku applicable for the billing tier with it). made it `per_machine_usage_cost` instead?

proto/viam/app/v1/billing.proto
```diff
@@ -71,6 +71,7 @@ message GetCurrentMonthUsageResponse {
   double discount_amount = 8;
   double total_usage_with_discount = 9;
   double total_usage_without_discount = 10;
+  double per_machine_usage_cost = 11;
```
ohEmily, line 74: I wonder if we should somehow indicate whether or not this field is in use for the current customer. There's a risk of ambiguity if someone has a 0 value for this (did not use their machines) versus being on a different model entirely.

So the question is should this billing.proto communicate what fields are in use given the current tier.

jr22: yeah agreed it could be ambiguous, but also true of any of the other usage counts in this API now. users can see the breakdown of costs in their invoice (along with the cost per unit). we could also add an a separate endpoint at some point that displays the billing tier information?

stevebriskin (approved):

PR #502, opened 2024-05-16
DATA-2469 Add framework/type to MLModelMetadata

katiepeters:


tahiyasalam:
Is there a corresponding app PR for this? We'll want to populate the model framework and model type information from the package metadata for ML Model, similar to how we do for the ML training scripts.

I also think we can revert [this commit](https://github.com/viamrobotics/api/pull/502/commits/821d0fa2c123a8c7635830d245a701f5420770f1) and instead use the `ready-for-protos` flag (more info [here](https://github.com/viamrobotics/api?tab=readme-ov-file#submitting-prs-for-protobuf-changes))

katiepeters:
@tahiyasalam there is not a corresponding app PR for this yet. Based on my limited understanding, it seemed like it made sense for this part to be done first? Or do they need to be done in tandem? Or..

tahiyasalam:
@katiepeters I would expect there to be an app (and if applicable RDK) PR created in tandem. This will allow us to stage changes to ensure that introducing the new proto field work as intended and that there are no breaking app if the message is being used somewhere, which in this case it is since MLModelMetadata exists already. If there are any changes that need to be made to the proto field(s) that we discover while coordinating the changes across repos, the change is as easy as a commit instead of a potentially breaking change to api.

tahiyasalam (approved):

PR #503, opened 2024-05-17
DATA-2630 - Add time to MLTrainingVersion

tahiyasalam:


agreenb (approved):

tahiyasalam:
@Simberific this is technically a breaking change for the submit custom training job API, though it has not currently yet been released. The change can be summarized as: the registry versions associated with training scripts will now have more metadata returned (the time created). Would you mind merging when you get a chance?

PR #504, opened 2024-05-17
Add tag to the emitted JSON blob

njooma:
Adding the tag version will allow the SDKs to pin to that specific release on buf.build

### DIFF
--- a/.github/workflows/tag_and_release_protos.yml
+++ b/.github/workflows/tag_and_release_protos.yml
@@ -47,7 +47,7 @@ jobs:
           tag: ${{needs.bump-tag.outputs.new_tag}}
 
   dispatch:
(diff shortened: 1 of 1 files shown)


abe-winter (approved):
lgtm

haven't looked at the receive side of this but I'm assuming an extra json field won't break anything

there is a [`toJSON` function](https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/expressions#tojson) in the actions language if you want to give it a shot, which would in theory allow you to do something like:

```yml
client-payload: toJSON({repo_name: github.repository, sha: github.sha, tag: needs.bump-tag.outputs.new_tag})
```

this is slightly safer in that it will do proper json escaping, but meh, your call

PR #508, opened 2024-05-24
Make fragment proto changes to accommodate unlisted fragments

piokasar:


DTCurrie:
For both messages, instead of deleting the `public` field and making a breaking change, should add a `[deprecated = true]` annotation or mark it `reserved` with a comment that points to `visibility`? Not sure what best practices are here.

So either:

```proto
optional bool public = 4 [deprecated=true];
// or
reserved 4; // Deprecated "public"
```

https://protobuf.dev/programming-guides/proto3/#reserved

piokasar:
When I've made breaking changes in the past, they have been merged but true, that probably was a special case. Maybe the API wasn't being used anywhere so doing breaking changes was fine. I added the `deprecated = true` option - eventually it looks like we can switch over to `reserved`. Thank you!

PR #509, opened 2024-05-29
RSDK-7745: Add SetRPM to API

martha-johnston:


proto/viam/component/motor/v1/motor.proto
```diff
@@ -43,6 +43,14 @@ service MotorService {
     };
   }
 
+  // SetRPM instructs the motor to move at the specified RPM indefinitely.
```
JohnN193, line 46: lets remove the comment from `GoFor` that talks about this behavior

JohnN193 (approved):

proto/viam/component/motor/v1/motor.proto
martha-johnston: I'll add a note that it's deprecated. in scope review yesterday we decided to leave the behavior as is for the next 3 ish months and mark it as deprecated in the meantime

JohnN193: o sounds good!

PR #516, opened 2024-06-06
APP-4949: TransferRegistryItem API

michaellee1019:
Adds protos for a TransferRegistryItem API allowing the transfer of ML models and modules from one organization to another.

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -195,6 +195,8 @@ service AppService {
 
   rpc DeleteRegistryItem(DeleteRegistryItemRequest) returns (DeleteRegistryItemResponse);
 
+  rpc TransferRegistryItem (TransferRegistryItemRequest) returns (TransferRegistryItemResponse);
+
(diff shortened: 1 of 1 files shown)


proto/viam/app/v1/app.proto
```diff
@@ -1007,6 +1009,13 @@ message DeleteRegistryItemRequest {
 
 message DeleteRegistryItemResponse {}
 
+message TransferRegistryItemRequest {
+  string item_id = 1;
+  string new_public_namespace = 2;
+}
+
+message TransferRegistryItemResponse {}
```
zaporter-work, line 1017: [aside] I wonder if we want to include the result URL or something for FE redirect? We can edit later to have that if we decide we want it.

zaporter-work (approved):
LGTM!

proto/viam/app/v1/app.proto
zaporter-work: Actually -- resolving this because it is not important for now.

piokasar (approved):

piokasar:
LGTM as well! Makes sense so far

PR #519, opened 2024-06-21
RSDK-7942: Add LogConfiguration field to ServiceConfig

danielbotros:
### Description
**[RSDK-7942:](https://viam.atlassian.net/browse/RSDK-7942)** Reconfiguring cloud robot's services' log levels does not propagate
* This bug occurs because the `ServiceConfig` message does not have a `LogConfiguration` field, so when converting proto objects to an actual config, the logging configuration is never properly set on a reconfiguration. 
* This PR adds a `LogConfiguration` field to the `ServiceConfig` message so that data is actually captured and can be used to propagate log level changes. 
* [This PR](https://github.com/viamrobotics/rdk/pull/4120) (WIP) makes the necessary changes to read and propagate log level changes


benjirewis (approved):
LGTM; let's wait for one other NetCode member to approve, though.

cheukt (approved):

PR #522, opened 2024-06-27
[APP-4842]: Add bson tags to data capture sensor readings types that are supported by tabular data ingested triggers.

maxhorowitz:
Hey!

Goal of this ticket is to add bson tags to the generated go structs for each of the 23 sensor readings data types that we plan to support for the triggers project. The reason for adding these tags is so that we can programmatically decode the datum that is uploaded to APP as part of validation that the sensor reading we've ingested is of the correct type for this trigger. This significantly improves the triggers code by protecting it against changes to the data types with which their logic is reliant. In the triggers code path, we will be importing the sensor types and unmarshalling readings into those types as follows:
```go
import (
	movSens "go.viam.com/api/component/movementsensor/v1"
)
```
```go
func validateMovementSensorLinearVelocityReading(datum bson.M) (bool, error) {
	bsonBytes, err := bson.Marshal(datum)
	if err != nil {
		return false, err
	}

	var res movSens.GetLinearVelocityResponse
	err = bson.Unmarshal(bsonBytes, &res)
	if err != nil {
		return false, err
	}
	...
}
```


proto/viam/component/movementsensor/v1/movementsensor.proto
```diff
@@ -113,7 +114,7 @@ message GetCompassHeadingRequest {
 message GetCompassHeadingResponse {
   // A number from 0-359 in degrees where
   // 0 is North, 90 is East, 180 is South, and 270 is   West
-  double value = 1;
+  double value = 1 [(tagger.v1.tags) = "bson:\"compass_heading\" json:\"compass_heading\""];
```
maxhorowitz, line 117:
```suggestion
  double value = 1 [(tagger.v1.tags) = "bson:\"value\" json:\"value\""];
```

proto/viam/common/v1/common.proto
```diff
@@ -113,8 +114,8 @@ message PointCloudObject {
 }
 
 message GeoPoint {
-  double latitude = 1;
-  double longitude = 2;
+  double latitude = 1 [(tagger.v1.tags) = "bson:\"latitude\""];
+  double longitude = 2 [(tagger.v1.tags) = "bson:\"longitude\""];
```
randhid, line 118: What happens if these ever change?

randhid (changes requested):
I think you're missing a few components, not sure if gripper's IsMoving is collected anywhere, it is a bool.

Also board's gpio pin and analog reader are collected I believe? How much coverage are you looking for?

proto/viam/common/v1/common.proto
maxhorowitz: @randhid Had a few discussions with some Core peeps and we're on the fence about whether to follow through with adding this validation altogether - it makes existing triggers reliant on the latest API types which is an issue for someone using a robot whose data format is out of sync with those latest types...I will circle back here if there are is need to add these tags, but for the meantime we will hold off on this addition

Thank you for hopping on this last week, sorry to have wasted any time

randhid: no worries! happy to jump in when/if the time comes.

PR #524, opened 2024-07-02
[APP-5400]: Add GetFragmentHistory proto

agavram:


piokasar:
ty! Added the request struct to the ticket as well, if we want to do it all in one pr then you could add it here : https://viam.atlassian.net/browse/APP-5400

piokasar (approved):

piokasar:
In the meeting with Michael Cousins, he said we should add a page length parameter to configure the number of documents returned. This should be an optional field called page_length. If not provided, the default should be 10.

PR #525, opened 2024-07-03
[APP-5420] Add ListMachineFragments request and response

piokasar:
https://viam.atlassian.net/browse/APP-5420

For reference in the tech spec,[ this ](https://docs.google.com/document/d/1SoTUIRQ8KpaZwcDirootBqBm-lclHg35uM9WSOGVSiA/edit#bookmark=id.tymrn9uczqrt)is why we are adding this api


proto/viam/app/v1/app.proto
```diff
@@ -774,6 +777,14 @@ message ListRobotsRequest {
   string location_id = 1;
 }
 
+message GetTopLevelAndNestedFragmentsRequest {
+  string part_id = 1;
```
mcous, line 781: Can we add a field to this request that allows the caller to specify additional fragment IDs? The user may have added more fragments locally but not yet saved, so the backend won't know about them yet, but it'd be great if the FE could stick to using a single endpoint to fetch all fragments relevant to the config

mcous (commented):
One request!

proto/viam/app/v1/app.proto
piokasar: ah yep - makes sense!

michaellee1019: I agree. I think instead of accepting `both` we only ask for a `repeated string fragment_ids` Then we can pass both saved and unsaved fragmentIds in one. I think it results in less confusing API behavior to have some fragmentIDs internally looked up from the part and other ones passed in.

mcous: Discussed in person, decided on:

- `string robot_id` - Allows the UI to the request all fragments in parallel with fetching the part configs, removing a request waterfall
- `repeated string fragment_ids` - Allows the UI to request additional fragments as the user adds them

michaellee1019: Agreed, I think to finalize this:

```
ListMachineFragmentsRequest {
 // the machine_id used to filter fragments defined in a machine's parts. Also returns any fragments nested within the fragments defined in parts.
 string machine_id = 1;
 // additional fragment_ids to append to the response. useful when needing to view fragments that will be provisionally added to the machine alongside existing fragments.
 repeated string additional_fragment_ids = 2;
}

message ListMachineFragmentsResponse{
  repeated Fragment fragments = 1;
}
```

piokasar: thanks both!

michaellee1019 (approved):

PR #527, opened 2024-07-03
RSDK-7899 Add MachineStatus to RobotService

maximpertsov:


proto/viam/robot/v1/robot.proto
```diff
@@ -360,3 +367,32 @@ message RestartModuleResponse {}
...
 
 message ShutdownResponse {}
+
+message GetMachineStatusRequest {}
+
+message GetMachineStatusResponse {
+  repeated ResourceStatus resources = 1;
+}
+
+message ResourceStatus {
```
maximpertsov, line 377: note that error is currently omitted - it will be added once we implement the "unhealthy" state

proto/viam/robot/v1/robot.proto
```diff
@@ -360,3 +367,33 @@ message RestartModuleResponse {}
...
+  enum State {
+    STATE_UNSPECIFIED = 0;
```
maximpertsov, line 379: this state is purely a default - it should never be returned by RDK

proto/viam/robot/v1/robot.proto
```diff
@@ -360,3 +367,33 @@ message RestartModuleResponse {}
...
 message ShutdownResponse {}
+
+message GetMachineStatusRequest {}
+
+message GetMachineStatusResponse {
+  repeated ResourceStatus resources = 1;
+}
+
+message ResourceStatus {
+  enum State {
```
benjirewis, line 378: [q] Does the defining the `enum` inside of `ResourceStatus` make any difference in the compiled code then if you were to define `enum` outside of the message?

proto/viam/robot/v1/robot.proto
```diff
@@ -360,3 +367,33 @@ message RestartModuleResponse {}
...
+    // a newly created resource.
+    STATE_UNCONFIGURED = 1;
+    // a resource that is being configured.
+    STATE_CONFIGURING = 2;
+    // a resource that has been successfully configured once and is not being
+    // re-configured or removed.
+    STATE_READY = 3;
+    // a resource that is being removed from the robot.
+    STATE_REMOVING = 4;
+  }
```
benjirewis, line 389: I wonder if adding `STATE_UNHEALTHY` eventually will be a breaking change. Not the end of the world, but curious how enum modifications are handled (i.e. if the "indices" of the other variants don't change, maybe it's not breaking).

proto/viam/robot/v1/robot.proto
```diff
@@ -142,6 +142,13 @@ service RobotService {
       post: "/viam/api/v1/shutdown"
     };
   }
+
+  // MachineStatus the current status of the robot.
```
benjirewis, line 146:
```suggestion
  // GetMachineStatus returns the current status of the robot.
```

proto/viam/robot/v1/robot.proto
maximpertsov: This shouldn't be breaking based on my read of https://protobuf.dev/programming-guides/enum/#definitions. 

The key point is that `proto3` treats enums as "open", which means that values outside of the enum range are set directly on the field. So an old client can receive messages with this field containing values that are outside of the defined enum range.

proto/viam/robot/v1/robot.proto
maximpertsov: It shouldn't based on https://github.com/bufbuild/buf/pull/400

> If the previous and current types are both enums, the enums are checked to see if the (1) the short names are equal (2) the previous enum is a subset of the current enum. A subset is defined as having a subset of the name/number enum values. If the previous is a subset, no failure is produced. The idea here is that this covers if someone just moves where an enum is defined, but still allows values to be added to this enum in the same change, as adding values to an enum is not a breaking change.

However, it's worth noting that this was not the case before: https://github.com/bufbuild/buf/issues/80

proto/viam/robot/v1/robot.proto
maximpertsov: which is great, since we might add new states beyond `unhealthy` in the future

proto/viam/robot/v1/robot.proto
benjirewis: Cool; thanks for looking into it!

proto/viam/robot/v1/robot.proto
benjirewis: Awesome, that's my reading, too.

benjirewis (approved):

component/audioinput/v1/audioinput.pb.gw.go
```diff
@@ -15,7 +15,7 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/utilities"
-	"go.viam.com/api/common/v1"
+	v1_0 "go.viam.com/api/common/v1"
```
dgottlieb, line 18: Not a blocking question, but do we know why this happened?

maximpertsov: my understanding is that the [formatter is unstable](https://viaminc.slack.com/archives/C01Q5S8V6TB/p1708015325914059?thread_ts=1708011768.501609&channel=C01Q5S8V6TB&message_ts=1708015325.914059)

dgottlieb: Got it, so this is just how your local code generator behaved. Thanks!

PR #531, opened 2024-07-16
RSDK-8228 Add config status to machine status API

maximpertsov:
Expand machine status API include config revision


proto/viam/robot/v1/robot.proto
```diff
@@ -395,5 +396,16 @@ message ResourceStatus {
   State state = 2;
 
   // state transition timestamp.
-  google.protobuf.Timestamp last_updated = 3;
```
maximpertsov, line 398: breaking change, but no systems are actually using this API yet

cheukt (approved):

PR #535, opened 2024-07-26
[APP-5417]: Accommodate pagination for GetRobotPartHistory in api

agavram:


piokasar (approved):
looks good! exact match with FragmentHistory makes sense

PR #537, opened 2024-07-31
RSDK-8381 add viam server version to API

ale7714:
https://viam.atlassian.net/browse/RSDK-8381


proto/viam/robot/v1/robot.proto
```diff
@@ -409,3 +416,15 @@ message ConfigStatus {
   // config ingestion timestamp.
   google.protobuf.Timestamp last_updated = 2;
 }
+
+message GetVersionRequest {}
+
+message GetVersionResponse {
+  // platform type of viam-server, either rdk or micro-rdk.
```
cheukt, line 423: don't know if we should specify that it will only be `rdk` or `micro-rdk`, since it may change

cheukt (approved):

PR #538, opened 2024-08-01
APP-5314 - add dates and limit to GetRobotPartLogsRequest

kim-mishra:
API Changes for [Improve Cloud Log Management](https://docs.google.com/document/d/1JBeAxRJ3igqYvf_zHRahPl4HrfPkvtZeT4jAsUCdw6E/edit#heading=h.tcicyojyqi6c)


proto/viam/app/v1/app.proto
```diff
@@ -640,6 +640,9 @@ message GetRobotPartLogsRequest {
   optional string page_token = 4;
   // logs of all levels are returned when the levels field is empty
   repeated string levels = 5;
+  optional google.protobuf.Timestamp start = 6;
+  optional google.protobuf.Timestamp end = 7;
+  optional int32 limit = 8;
```
jr22, line 645: any rhyme or reason that you can discern why we use `int32` vs `int64` in some places?

proto/viam/app/v1/app.proto
```diff
@@ -640,6 +640,9 @@ message GetRobotPartLogsRequest {
   optional string page_token = 4;
   // logs of all levels are returned when the levels field is empty
   repeated string levels = 5;
+  optional google.protobuf.Timestamp start = 6;
+  optional google.protobuf.Timestamp end = 7;
```
jr22, line 644: i thought we decided to not make thse optional?

proto/viam/app/v1/app.proto
kim-mishra: no honestly, i cmd+f and saw int32 so I used it. Just took a look and still can't discern any patternbut it should probably be an int64 right? unless you think there is any con to having int64 over int32

proto/viam/app/v1/app.proto
jr22: ah ok great! thanks for clarifying

jr22 (approved):

PR #539, opened 2024-08-02
RSDK-7903 Add unhealthy state and error field

maximpertsov:
Add the resource unhealthy state + field detailing what error caused the resource to be unhealthy.

Related RDK change:
* https://github.com/viamrobotics/rdk/pull/4257


proto/viam/robot/v1/robot.proto
```diff
@@ -389,11 +389,13 @@ message ResourceStatus {
     STATE_UNCONFIGURED = 1;
     // a resource that is being configured.
     STATE_CONFIGURING = 2;
-    // a resource that has been successfully configured once and is not being
+    // a resource that has been successfully configured once, is healthy, and is not being
```
ale7714, line 392: does healthy mean ready state? maybe we can explicitly say ready for clarity

ale7714 (approved):

proto/viam/robot/v1/robot.proto
maximpertsov: Yeah this is not as clear as it could be - I changed this to "not unhealthy" since we have a definition for that state. I don't want to conflate "ready" with "healthy" too much since a resource can be "healthy" (no errors) but re-configuring or removing.

PR #540, opened 2024-08-09
add last updated to fragment proto

puranamp:
needed for [this](https://github.com/viamrobotics/app/pull/5679)


maximpertsov (approved):

michaellee1019 (approved):
Fine with this change. Technically you could achieve what you need to do in your project without any API changes. Internally can store the `last_updated` field, and return that field as part of the config response to the robot as the config version. Doing this API work makes the APIs match what is stored on the Fragment, so this is good practice.

I can send out this API change as an email to the scope, any API change made is supposed to be in a scope/email.

michaellee1019:
I just looked at other protos. Don't we have the same issue for RobotPart and Location as well? Wondering if we want to expand this PR to handle those and do it all at once.

I came to realize this when reviewing the naming of other fields (one usage named `last_updated` and another `updated_at`).

michaellee1019 (approved):
LGTM! I suggest adding a comment for the robot part field. Comments show up in editors when you are interfacing with the proto objects.

I sent an email mentioning the proto change to scope reviewers.

PR #544, opened 2024-08-13
APP-3604 build endpoints for external oauth and repos

abe-winter:
## What changed
- build service endpoints for:
    - external oauth app links
    - and for module-to-repo links
## Why
Linking our build service to source control / CI is currently bad; it's super unintuitive and doesn't work with private repos. These endpoints let us use native oauth tools provided by git hosts for something closer to one-click setup. (no more yamls! no need to set up viam credentials).

Note the new endpoints are experimental (the entire build service is still marked as 'subject to change'). They don't work yet and should only be hit by the viam CLI, not external users.


proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,74 @@ message ListJobsResponse {
...
+  // OrgID of the module
+  string org_id = 2;
+  // public namespace of the module
+  optional string namespace = 3;
+  // ID of the module (just the name, not the dotted org:module form)
+  string module_id = 4;
+  // git repo in owner/repository form
+  string repo = 5;
+  // email of the viam user who created this
+  string viam_user = 6;
```
ale7714, line 114: Will it be an issue if the viam user is deleted?

proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,74 @@ message ListJobsResponse {
   // jobs is ordered by (build start time, alphabetical platform).
   repeated JobInfo jobs = 1;
 }
+
+message RepoLink {
+  string app_link_id = 1;
```
ale7714, line 104: Is this an Id from GH?

ale7714: if it is, can we call oauth_app_link id or something? so it's a bit more clear where it comes from

abe-winter: yup way clearer, done in 33c41a0

ale7714 (approved):
🚀

proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,79 @@ message ListJobsResponse {
...
+
+message RepoLink {
+  // viam internal ID which maps to an external oauth app
+  string oauth_app_link_id = 1;
+  // OrgID of the module
+  string org_id = 2;
+  // public namespace of the module
+  optional string namespace = 3;
+  // ID of the module (just the name, not the dotted org:module form)
+  string module_id = 4;
```
zaporter-work, line 111: Should we call this module_name?

We use module_id pretty ubiquitously to mean {org|namespace}:name

proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,79 @@ message ListJobsResponse {
...
+  // git repo in owner/repository form
+  string repo = 5;
```
zaporter-work, line 113:
```suggestion
  // github repo in owner/repository form
  string repo = 5;
```
?

proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,79 @@ message ListJobsResponse {
...
+// represents a link between viam users / orgs and an external oauth app
+message AppLink {
+  // viam internal ID which maps to an external oauth app
+  string oauth_app_link_id = 1;
+  // email of the viam user that owns this
+  string viam_user = 2;
+  // username of the external user who created this
+  string external_user = 3;
+  // list of org public namespace (where available) or org UUIDs attached to the external app
+  repeated string org_id_or_ns = 4;
```
zaporter-work, line 149: To support one user attaching multiple orgs to the same app?

proto/viam/app/build/v1/build.proto
```diff
@@ -83,3 +99,79 @@ message ListJobsResponse {
...
   repeated JobInfo jobs = 1;
 }
+
+message RepoLink {
+  // viam internal ID which maps to an external oauth app
+  string oauth_app_link_id = 1;
+  // OrgID of the module
+  string org_id = 2;
+  // public namespace of the module
+  optional string namespace = 3;
```
zaporter-work, line 109: [probably worth discussing] Why is associating the namespace necessary for this message (same for AppLink)?

I worry that this will mean that will add a lot more logic & edge cases for module transfers from org to org.

zaporter-work (approved):
I have low context, but in isolation, these seem good

proto/viam/app/build/v1/build.proto
abe-winter: yes -- main motivation for this was our internal use pattern

proto/viam/app/build/v1/build.proto
abe-winter: this is populated in the downstream direction so that CLI output is user-readable; it's not stored in the DB and it's not used in the request. I'll edit the docstring to say this

proto/viam/app/build/v1/build.proto
abe-winter: done in d6cafd9

proto/viam/app/build/v1/build.proto
abe-winter: felt uncomfortable using a vendor name in our api; and also this could hypothetically be used in other places in the future

proto/viam/app/build/v1/build.proto
zaporter-work: SGTM

proto/viam/app/build/v1/build.proto
abe-winter: this is just for tracking history; the authorization is done through the OAuth app link, which is permissioned per viam org

PR #550, opened 2024-09-06
DATA-3094: Add CLI args to SubmitCustomTrainingJobRequest

vpandiarajan20:
Adds `argument` field to SubmitCustomTrainingJobRequest, map of string to string.


proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -69,6 +69,7 @@ message SubmitCustomTrainingJobRequest {
   string organization_id = 3 [(tagger.v1.tags) = "bson:\"organization_id\" json:\"organization_id\""];
   string model_name = 4 [(tagger.v1.tags) = "bson:\"model_name\" json:\"model_name\""];
   string model_version = 5 [(tagger.v1.tags) = "bson:\"model_version\" json:\"model_version\""];
+  map<string, string> arguments = 6;
```
tahiyasalam: Do we needs tags for this cc @sbal13 ?

tahiyasalam (approved):

proto/viam/app/mltraining/v1/ml_training.proto
sbal13: Sorry not sure what you mean. As in tags like we have for `SubmitTrainingJobRequest`?

```
  repeated string tags = 6 [(tagger.v1.tags) = "bson:\"tags\" json:\"tags\""];
```

If so, then no, we don't have UI for selecting what tags to train on - custom scripts don't take tags IIRC, only the built-in script does. Do you think it would be useful to add?

tahiyasalam: I meant the bson and json tags, though I'm not exactly sure what they're used for in this scenario.

sbal13: Yeah I'm not sure either! It looks like they're used to tell protobuf what attribute name to map to when marshalling (I have no idea if I'm using that word correctly, converting??) to BSON and JSON respectively, but in all cases the attribute name is the same in this file so not sure what it's doing.

PR #552, opened 2024-09-12
[APP-6230]: Add proto definition of RPC for SendEmailForBillingTierPayment.

maxhorowitz:


proto/viam/app/v1/billing.proto
```diff
@@ -109,3 +111,13 @@ message GetInvoicePdfRequest {
 message GetInvoicePdfResponse {
   bytes chunk = 1;
 }
+
+message SendEmailForBillingTierPaymentRequest {
+  google.protobuf.Timestamp request_time = 1;
+  string org_id = 2;
+  string email = 3;
```
maxhorowitz, line 118: Do you think we want this to be email or user ID? In our prod db we don't even have one user with multiple emails, so they're effectively 1<>1. From perspective of client, I'm pretty sure any of the SDKs would have access to either of these things?

Or --- do we omit the email/userID entirely and rely on the user's state (which gets populated in the middleware before serving this endpoint)?

PR #557, opened 2024-09-23
RSDK-8716 - add config message to robot proto

Kschappacher:
## Feature

- define maintenance sensor messages
- Add messages to the robot config


cheukt (approved):

cheukt:
the ticket should be RSDK-8716, can you update before merging?

PR #559, opened 2024-09-30
DATA-3210 - Add MLInference service and GetInference RPC to proto

tahiyasalam:
### Context
This PR introduces the new ML Inference service and GetInference request. I am cautious about including the response and would like to leave that to a future ticket in case anything changes. What do you think? 

Here is the [app PR](https://github.com/viamrobotics/app/pull/6327) with the changes staged.


etai-shuchatowitz (approved):
This makes sense to me. I don't see us getting to the response any time over the next 2 weeks so happy to leave this unimplemented for the time being.

PR #563, opened 2024-10-09
APP-6580: Adding fragment + org cost structure

RoxyFarhad:
Our current `GetCurrentMonthUsage` response is built only for Org related costs. We now want to show fragment charges + credits via this API as well as org level costs (which we do because we use it in [APP](https://github.com/viamrobotics/app/pull/6451)) then we need to update how we return this information. 

This change allows us to group by orgs and fragments and show how the discounts are applied to each of these separate charge types. 

**As as drive by** here I removed the types `Invoice` and `BillableResourceEvent` as they are not used anywhere and is confusing to people.

### DIFF
--- a/proto/viam/app/v1/billing.proto
+++ b/proto/viam/app/v1/billing.proto
@@ -28,16 +28,6 @@ message InvoiceSummary {
   google.protobuf.Timestamp paid_date = 6;
 }
 
-message BillableResourceEvent {
-  string id = 1;
-  string type = 2;
-  double usage_quantity = 3;
-  string usage_quantity_unit = 4;
-  string usage_cost = 5;
-  google.protobuf.Timestamp occurred_at = 6;
-  string user_name = 7;
-}
-
 enum PaymentMethodType {
   PAYMENT_METHOD_TYPE_UNSPECIFIED = 0;
   PAYMENT_METHOD_TYPE_CARD = 1;
--- a/proto/viam/app/v1/billing.proto
+++ b/proto/viam/app/v1/billing.proto
@@ -38,16 +38,6 @@ message BillableResourceEvent {
   string user_name = 7;
 }
 
-message Invoice {
-  string id = 1;
-  google.protobuf.Timestamp invoice_date = 2;
-  double invoice_amount = 3;
-  string status = 4;
-  google.protobuf.Timestamp due_date = 5;
-  repeated BillableResourceEvent items = 6;
-  string emailed_to = 7;
-}
-
 enum PaymentMethodType {
   PAYMENT_METHOD_TYPE_UNSPECIFIED = 0;
   PAYMENT_METHOD_TYPE_CARD = 1;
@@ -74,9 +64,21 @@ enum UsageCostType {
   USAGE_COST_TYPE_PER_MACHINE = 8;
 }
 
+enum SourceType {
+  SOURCE_TYPE_UNSPECIFIED = 0;
+  SOURCE_TYPE_ORG = 1;
+  SOURCE_TYPE_FRAGMENT = 2;
+}
+
 message UsageCost {
   UsageCostType resource_type = 1;
   double cost = 2;
+}
+
+message ResourceUsageCostsBySource {
+  SourceType source_type = 1;
+  ResourceUsageCosts resource_usage_costs = 2;
+  string tier_name = 3;
 }
 
 message ResourceUsageCosts {
@@ -86,18 +88,11 @@ message ResourceUsageCosts {
   double total_without_discount = 4;
 }
 
-message FragmentUsageCosts {
-  string fragment_id = 1;
-  ResourceUsageCosts resource_usage_costs = 2;
-}
-
 message GetCurrentMonthUsageResponse {
   google.protobuf.Timestamp start_date = 1;
   google.protobuf.Timestamp end_date = 2;
-
-  ResourceUsageCosts org_usage = 14;
-  repeated FragmentUsageCosts fragment_usage_costs = 15;
-  double subtotal = 16;
+  repeated ResourceUsageCostsBySource resource_usage_costs_by_source = 3;
+  double subtotal = 4;
 }
 
 message GetOrgBillingInformationRequest {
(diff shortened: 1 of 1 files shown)


proto/viam/app/v1/billing.proto
```diff
@@ -62,9 +62,39 @@ message GetCurrentMonthUsageRequest {
...
+  DATA_UPLOAD = 2; 
+  DATA_EGRESS = 3;
+  REMOTE_CONTROL = 4; 
+  STANDARD_COMPUTE = 5; 
+  BINARY_DATA_CLOUD_STORAGE = 6; 
+  OTHER_CLOUD_STORAGE = 7;
+  PER_MACHINE = 8;
+}
+
+message UsageCost {
```
RoxyFarhad, line 76: this is to replace the old way of how we did this which was just list the costs directly on the struct. We want something like this so its usable for fragments too

proto/viam/app/v1/billing.proto
```diff
@@ -62,9 +62,40 @@ message GetCurrentMonthUsageRequest {
...
+  string fragment_id = 1;
+  GroupedCosts breakdown = 2;
+}
+
 message GetCurrentMonthUsageResponse {
   google.protobuf.Timestamp start_date = 1;
   google.protobuf.Timestamp end_date = 2;
+
+  // This will be deprecated soon as it no longer
+  // reflects fragment cost structure
```
ohEmily, line 98: [q] why can't we deprecate now?

proto/viam/app/v1/billing.proto
```diff
@@ -76,6 +107,10 @@ message GetCurrentMonthUsageResponse {
   double per_machine_usage_cost = 11;
   double binary_data_cloud_storage_usage_cost = 12;
   double other_cloud_storage_usage_cost = 13;
+
+  repeated GroupedCosts org_breakdown = 14;
+  repeated FragmentCosts fragment_breakdown = 15;
```
ohEmily, line 112: optional?

proto/viam/app/v1/billing.proto
```diff
@@ -62,9 +62,40 @@ message GetCurrentMonthUsageRequest {
...
+  USAGE_COST_TYPE_OTHER_CLOUD_STORAGE = 7;
+  USAGE_COST_TYPE_PER_MACHINE = 8;
+}
+
+message UsageCost {
+  UsageCostType resource_type = 1;
+  double cost = 2;
+}
+
+message GroupedCosts {
```
ohEmily, line 82: `SumTotalCosts`?

ohEmily: q: why can there be multiple `GroupedCosts`? i.e. repeated

ohEmily: A: will update to single per org

proto/viam/app/v1/billing.proto
```diff
@@ -62,9 +62,40 @@ message GetCurrentMonthUsageRequest {
...
+  repeated UsageCost usage_costs = 1;
+  double subtotal = 2;
```
ohEmily, line 84: `totalNoDiscount`?

proto/viam/app/v1/billing.proto
```diff
@@ -62,20 +62,42 @@ message GetCurrentMonthUsageRequest {
...
-  double standard_compute_usage_cost = 7;
-  double discount_amount = 8;
-  double total_usage_with_discount = 9;
-  double total_usage_without_discount = 10;
-  double per_machine_usage_cost = 11;
-  double binary_data_cloud_storage_usage_cost = 12;
-  double other_cloud_storage_usage_cost = 13;
+
+  ResourceUsageCosts org_usage = 14;
+  repeated FragmentUsageCosts fragment_usage_costs = 15;
```
jr22, line 99: instead of making these so special cased with terminology such as "org" and "fragment", can we combine them? ultimately what we want to do here is support multiple billing tiers per org right? would be nice if this were an array of billing tiers + costs instead of two distinct separate variables and you could add a boolean like "defaultTier" to indicate which one is the org tier. picturing something like this (ignore my bad names) 

```
message BreakdownByBillingTiers {
  string tier_name = 1; 
  boolean is_default_tier = 2
  GroupedCosts ResourceUsageCosts = 3;
}

repeated BreakdownByBillingTiers breakdownByTier = 14;
```

i think this makes this API more flexible in the future. what do you think?

jr22 (commented):
meta comment-- there are probably other billing APIs that we'd need to fix too to align with the concept that an org can have multiple tiers + have different credits. i feel ok punting that to https://viam.atlassian.net/browse/APP-5943 where we are going to be thinking about this architecture more but wanted to call out that there are other APIs with this issue

proto/viam/app/v1/billing.proto
jr22: great! looks much cleaner IMO. thanks!

RoxyFarhad:
> meta comment-- there are probably other billing APIs that we'd need to fix too to align with the concept that an org can have multiple tiers + have different credits. i feel ok punting that to https://viam.atlassian.net/browse/APP-5943 where we are going to be thinking about this architecture more but wanted to call out that there are other APIs with this issue

I am not actually sure whether we will need to change anything else, the other message definitions that seem relevant were not used so I am removing

jr22:
> > meta comment-- there are probably other billing APIs that we'd need to fix too to align with the concept that an org can have multiple tiers + have different credits. i feel ok punting that to https://viam.atlassian.net/browse/APP-5943 where we are going to be thinking about this architecture more but wanted to call out that there are other APIs with this issue
> 
> I am not actually sure whether we will need to change anything else, the other message definitions that seem relevant were not used so I am removing

oh gotcha. i think the other places im thinking of are internal APIs which should be straight forward then

proto/viam/app/v1/billing.proto
```diff
@@ -28,26 +28,6 @@ message InvoiceSummary {
   google.protobuf.Timestamp paid_date = 6;
 }
 
-message BillableResourceEvent {
```
RoxyFarhad, line 31: These should still be removed as they are not used anywhere and have never been so it doesn't make sense to even "deprecate" them

ohEmily (approved):

PR #565, opened 2024-10-10
add `StartBuildRequest.workdir` field

abe-winter:
## What changed
- `workdir` field
## Why
Support repos with multiple modules.


bhaney:
Nice

bhaney (approved):

PR #567, opened 2024-10-16
APP-6392: add errors field to ListMachineFragments response

mcous:
## Overview

This PR adds the necessary proto fields to report nested fragments errors from the `ListMachineFragments` endpoint

## Change log

- Define `FragmentError` message
- Add `repeated FragmentError` to `ListMachineFragmentsResponse`

Note: APP-6392 spec'd adding errors to the `Fragment` object. After getting into the code, I'm not sure that's the best move. Will probably keep this PR in draft for a while so I can spin up APP-5422 and see how it feels from the top-down rather than try to get this shape right from the bottom up

## Review requests

Check back later!


mcous:
Closing in favor of #576

PR #574, opened 2024-10-24
RSDK-9125 add MoveThroughJointPositions proto

raybjork:
This PR adds the proto described by the scope doc https://docs.google.com/document/d/1Vbbt6wnILJ1CYx-sQWcow8pzIBsFLfgV6BLhfZKRfR0/edit


biotinker (approved):
LGTM

proto/viam/component/arm/v1/arm.proto
```diff
@@ -43,6 +43,15 @@ service ArmService {
     };
   }
 
+  // MoveThroughJointPositions moves every joint on a robot's arm to the specified JointPositions in the order they are specified, obeying the specified velocity and acceleration limits.
+  // This will block until done or a new operation cancels this one
+  rpc MoveThroughJointPositions(MoveThroughJointPositionsRequest) returns (MoveThroughJointPositionsResponse) {
+    option (common.v1.safety_heartbeat_monitored) = true;
+    option (google.api.http) = {
+      put: "/viam/api/v1/component/arm/{name}/joint_positions"
```
raybjork, line 51: This should have come up in the scope but this is also the path we're using for the other joint positions call and might need to change?

biotinker: Oh lol yeah that won't work.

Call it `repeated_joint_positions` maybe?

raybjork: I just went with a snake case version of the method name for maximum clarity

biotinker: Makes sense, looks good

PR #576, opened 2024-10-24
APP-6392: Update fragment protos 

anjinai:
## Overview

Updates list machine fragment response to return fragment trees that contain a root id, a merge order, and fragment errors for that tree.


proto/viam/app/v1/app.proto
```diff
@@ -745,6 +745,26 @@ enum FragmentVisibility {
...
+  string fragment_id = 2;
+  string detail = 3;
+}
+
+enum FragmentErrorType {
+  FRAGMENT_ERROR_TYPE_UNSPECIFIED = 0;
+  FRAGMENT_ERROR_TYPE_NO_ACCESS = 1;
+  FRAGMENT_ERROR_TYPE_NESTING_LIMIT_EXCEEDED = 2;
+  FRAGMENT_ERROR_TYPE_CHILD_ID_INVALID = 3;
+  FRAGMENT_ERROR_TYPE_DUPLICATE_NAMES = 4;
```
mcous, line 759: Let's leave this error type out for now; we can add it in when we actually add the functionality. Might be a bit much to deal with at the moment

proto/viam/app/v1/app.proto
```diff
@@ -745,6 +745,25 @@ enum FragmentVisibility {
   FRAGMENT_VISIBILITY_PUBLIC_UNLISTED = 3;
 }
 
+message FragmentError {
+  FragmentErrorType error_type = 1;
+  string fragment_id = 2;
+  string detail = 3;
```
ehhong, line 751: should this be `message`?

ehhong (approved):

proto/viam/app/v1/app.proto
```diff
@@ -745,6 +745,25 @@ enum FragmentVisibility {
...
+}
+
+enum FragmentErrorType {
+  FRAGMENT_ERROR_TYPE_UNSPECIFIED = 0;
+  FRAGMENT_ERROR_TYPE_NO_ACCESS = 1;
+  FRAGMENT_ERROR_TYPE_NESTING_LIMIT_EXCEEDED = 2;
+  FRAGMENT_ERROR_TYPE_CHILD_ID_INVALID = 3;
```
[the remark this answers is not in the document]
anjinai: Good call, I'm going to go with `FRAGMENT_ERROR_TYPE_CYCLE_DETECTED = 4` unless you have any objections.

proto/viam/app/v1/app.proto
anjinai: I think it makes sense to be a message unless I'm missing something.

mcous: Picked `detail` just because I saw that in other protos that represent errors, no strong preference though

PR #578, opened 2024-10-25
APP-6752: Add 'fragmentContents' and deprecate 'old' in FragmentHistoryEvent

ohEmily:
[APP-6752](https://viam.atlassian.net/browse/APP-6752)

We do not want docs in the `fragment_history` collection to include metadata that is shared with the doc in the `fragment` collection. These duplicated fields are creating a bug in the UI and allowing rollbacks to versions on no-op changes made to immutable fields.


ohEmily:
iteration note to self: `make up-backend` chokes on these changes as follows:
```
data/data.go:2305:21: cannot use *old (variable of type Fragment) as FragmentContents value in struct literal
data/models.go:1076:13: cannot use oldFragmentContentsProto (variable of type *structpb.Struct) as *"go.viam.com/api/app/v1".Fragment value in struct literal
data/test_helpers.go:278:40: unknown field Fragment in struct literal of type Fragment
data/test_helpers.go:828:3: unknown field Fragment in struct literal of type Fragment
make[1]: *** [up-app-local-db] Error 1
make[1]: *** Waiting for unfinished jobs....
```

dgottlieb:
@ohEmily @michaellee1019 can this PR be closed? It seems like we at least did [some portion of it](https://github.com/viamrobotics/api/pull/622) earlier this year.

PR #579, opened 2024-10-31
APP-6794: Add APIs for white-labelled invoicing

RoxyFarhad:


proto/viam/app/v1/app.proto
```diff
@@ -61,6 +61,18 @@ service AppService {
   // Resend an organization invite
   rpc ResendOrganizationInvite(ResendOrganizationInviteRequest) returns (ResendOrganizationInviteResponse);
 
+  // Third party
+
+  rpc EnableBillingService(EnableBillingServiceRequest) returns (EnableBillingServiceResponse);
+
+  rpc DisableBillingService(DisableBillingServiceRequest) returns (DisableBillingServiceResponse);
+
+  rpc UpdateBillingService(UpdateBillingServiceRequest) returns (UpdateBillingServiceResponse);
```
jr22, line 70: do we want these in the app protos?

RoxyFarhad: I also went back and forth on this, because its related to the "org" and not billing itself, I thought it made more sense in the app proto. Open to discussion though

jr22: i think @maxhorowitz had some thoughts on if we wanted a separate "third party service" service

RoxyFarhad: open to discussion but I think that too complex for what is 5 new RPCs right now. I think that this should follow a similar pattern to the CLI design where its a org-level thing (which lives on app).

jr22: yeah looking again at the CLI commands-- since theyre `viam organization billing-service ... ` and not `viam third-party billing-service --org-id aoisjfdoiasjdf` (or somethign lke that) i agree where it is now makes sense

proto/viam/app/v1/app.proto
```diff
@@ -411,6 +423,51 @@ message DeleteOrganizationMemberRequest {
...
+
+message OrganizationGetSupportEmailRequest {
+  string org_id = 1;
+}
+
+message OrganizationGetSupportEmailResponse {
+  string email = 1;
+}
+
+message EnableBillingServiceResponse {}
```
jr22, line 455: sorry this is soooo nitty but can we move this up so its paired with its request? all the others are paired

jr22 (approved):
something i just thought of is we probs also want a "get" for the billing service config stuff  but can always add later if you want

RoxyFarhad:
> something i just thought of is we probs also want a "get" for the billing service config stuff but can always add later if you want

oh that's probably true. I think that will require a scope change though, so I am going to make a ticket for it but punt on it

maxhorowitz (approved):

PR #580, opened 2024-11-01
Add username to ProcessConfig proto

zaporter-work:
See comments here: https://github.com/viamrobotics/docs/pull/3649

Looks like [username](https://github.com/viamrobotics/goutils/pull/206) was never actually added to the protos -- so it doesn't work in the UI.

I'm not sure if you ever expected this to be used in the UI.

@abe-winter We have a few options from here:
- Don't document username (not sure if it was ever supposed to be used from the app config)
- Add username to the protos (Are you hesitant about adding to protos?)


abe-winter (approved):
lgtm

for clarity -- this only works with processes, not with modules? (or do modules already have a way to do it?)

zaporter-work:
Oops -- forgot to add @cheukt for code-owner approval.

PR #582, opened 2024-11-04
APP-6290 Update list machine fragments proto to return resolved fragments

ehhong:
### Overview
This PR changes the shape of the `ListMachineFragments` response to return the fully resolved fragment configs (resolving nesting) instead of the previous ambiguous "fragment tree" and "merge order." This was proposed after discussion with @jr22 and @anjinai and aims to improve the clarity of this endpoint.


proto/viam/app/v1/app.proto
```diff
@@ -895,7 +895,7 @@ message ListMachineFragmentsRequest {
 
 message ListMachineFragmentsResponse {
   repeated Fragment fragments = 1;
-  repeated FragmentTree fragment_trees = 2;
+  repeated FragmentWithError resolved_fragments = 2;
```
ehhong, line 898: fyi, this has some implications for nested fragments and fragment overwrites behavior. if we return the resolved fragments (the resultant config after all nesting is applied), fragment overwrites will operate on the resolved fragment as a whole, rather than on each nested fragment within a fragment tree.

here is an example. given this fragment tree:
```
a
|
b
|
c
```
if i want to apply a fragment overwrite to fragment b (nested within fragment a), i add a fragment mod to fragment a in my part config.

the alternative to this is to have fragment mods apply to nested fragments. so, given the same fragment tree, you would include a fragment mod on fragment b in the part config. this would allow you to resolve any conflicts between nested fragments (like fragments b and c) before they are merged. however, i think this approach is confusing to use because it requires the user to understand the nested structure of the fragment, and this would require us to push more duplicate computation on the frontend. @mcous and i spoke and we think it's okay to push nested-fragment-resolution to a later project (ie. fragment overwrites on fragments). 

please leave any thoughts you have on this!

proto/viam/app/v1/app.proto
```diff
@@ -816,9 +816,9 @@ enum FragmentErrorType {
   FRAGMENT_ERROR_TYPE_CYCLE_DETECTED = 4;
 }
 
-message FragmentWithError {
+message ResolvedFragment {
```
jr22, line 819: dont have a strong opinion on terminology but floating some synonyms to resolved to make sure we feel good about it: "expanded" or "flattened". i still like "resolved" but figured id toss some out to see how we feel

jr22 (approved):
thanks emily!

proto/viam/app/v1/app.proto
jr22: great i agree that pushing the resolution to a later project works as long as this decision isnt going to make that _more_ difficult. this current logic still makes sense to me too

proto/viam/app/v1/app.proto
ehhong: ooo thank you for the suggestions, i think i still prefer "resolved" because it leaves us some space in case computing fragments means something else in the future

mcous (approved):
LGTM

anjinai (approved):

ehhong:
@stevebriskin this is a breaking change for an in-flight feature (nested fragments). the updated fields are not being used anywhere except internally, so shouldn't affect anyone to break them

PR #583, opened 2024-11-05
APP-6158: Remove billing support email in favor of shared support email between billing and auth

jr22:
Smol tweak that came out of APP-6158-- we no longer need `billing_support_email` on the `UpdateBillingServiceRequest` since we will have a shared API for setting/getting the support email (defined here: https://github.com/viamrobotics/api/pull/579/files#diff-e842ca0f50219c67d40be90a2a1efd40ee1b6df19a254357ab177d631b3ea4b4R72-R74)


RoxyFarhad (approved):

proto/viam/app/v1/app.proto
```diff
@@ -443,7 +443,6 @@ message EnableBillingServiceResponse {}
 message UpdateBillingServiceRequest {
   string org_id = 1;
   BillingAddress billing_address = 2;
-  string billing_support_email = 3;
```
jr22, line 446: @ohEmily could you force merge this? removed this field because we realized we no longer need it (have separate email api since shared with auth service). currently unused so not really a breaking change

PR #592, opened 2024-11-14
[APP-6858]: Add all RPC API definitions to AppService that will be required as part of the Self-Service FusionAuth epic.

maxhorowitz:
https://viam.atlassian.net/browse/APP-6858
Screenshot 2024-11-14 at 2 09 55 PM


proto/viam/app/v1/app.proto
```diff
@@ -1411,3 +1423,77 @@ message CreateKeyFromExistingKeyAuthorizationsResponse {
...
+message ReadOAuthAppResponse {
+  string name = 1;
+  string secret = 2;
+  OAuthConfig oauth_config = 3;
+}
+
+message UpdateOAuthAppRequest {
+  string org_id = 1;
+  string id = 2;
+  string name = 3;
```
jr22, line 1473: application_id and application_name?

proto/viam/app/v1/app.proto
```diff
@@ -1411,3 +1423,77 @@ message CreateKeyFromExistingKeyAuthorizationsResponse {
...
+  string org_id = 1;
+}
+message ListOAuthAppsResponse {
+  repeated string ids = 1;
+}
+
+message OAuthConfig {
+  string client_authentication = 1;
+  string pkce = 2;
+  string url_validation = 3;
```
jr22, line 1494: what are example values of these? for my own knowledge

proto/viam/app/v1/app.proto
```diff
@@ -75,6 +75,18 @@ service AppService {
...
+  rpc OrganizationSetLogo(OrganizationSetLogoRequest) returns (OrganizationSetLogoResponse);
+  rpc OrganizationGetLogo(OrganizationGetLogoRequest) returns (OrganizationGetLogoResponse);
+
+  rpc EnableAuthService(EnableAuthServiceRequest) returns (EnableAuthServiceResponse);
+  rpc DisableAuthService(DisableAuthServiceRequest) returns (DisableAuthServiceResponse);
+
+  rpc CreateOAuthApp(CreateOAuthAppRequest) returns (CreateOAuthAppResponse);
+  rpc ReadOAuthApp(ReadOAuthAppRequest) returns (ReadOAuthAppResponse);
+  rpc UpdateOAuthApp(UpdateOAuthAppRequest) returns (UpdateOAuthAppResponse);
+  rpc DeleteOAuthApp(DeleteOAuthAppRequest) returns (DeleteOAuthAppResponse);
```
jr22, line 87: fwiw we explicitly decided _not_ to implement delete in the original scope of work. this is because delete is a non-reversible and pretty large impact operation so we dont want anyone doing it by mistake. i know you want to implement all CRUD operations but 🤷‍♀️ . i dont think having to reach out to support to delete an app is an antipattern.

jr22 (commented):
food for thought though for future planning: i usually find it more intuitive to implement the API for each alongside the relevant API changes instead of batching the API changes like this. extra params often come up once you start implementing or you realize a field is unnecessary. much easier to do the API design once you have the implementation. can always change these later though if that happens. 

when are you planning to delete the old apis? once the new ones are implemented? do you have a ticket for deprecation + removing

proto/viam/app/v1/app.proto
maxhorowitz: My opinion is anyone who is savvy enough to be using a CLI to generate OAuth apps that work have sufficient knowledge to be deleting them ... maybe we can implement a two step safety solution where the user has to double confirm that they want to delete?

maxhorowitz:
@jr22 added placeholder ticket for removal of existing APIs here: https://viam.atlassian.net/browse/APP-6993

proto/viam/app/v1/app.proto
maxhorowitz: `client_authentication` can be one of "Required", "NotRequired", or "NotRequiredWhenUsingPKCE"
`pkce` can be one of "Required", "NotRequired", or "NotRequiredWhenUsingClientAuthentication"
`url_validation` can be one of "ExactMatch" or "AllowWildcards"

(... those strings are exactly how its defined in the FusionAuth API, but maybe we can expose it to the user a little differently and match to the FusionAuth values we will need ... )

proto/viam/app/v1/app.proto
maxhorowitz: Sure -- whats the deal with this stuff considering it was approved in scope that way?

maxhorowitz: (I'm sure fine but asking to be safe)

proto/viam/app/v1/app.proto
jr22: should we make enums for that?

proto/viam/app/v1/app.proto
jr22: its not about being savvy enough, its more that its a big operation that we dont want people to do lightly. agreed that in the CLI could be good to implement that prompt to confirm.

proto/viam/app/v1/app.proto
jr22: ive been told that its fine to tweak api that was posed in scope as long as its minimal

proto/viam/app/v1/app.proto
maxhorowitz: Yes we can good idea!

proto/viam/app/v1/app.proto
maxhorowitz: In practice its a client ID, which will make the most sense to a customer who is trying to debug their auth flows because they will see the client ID in their browser's nav bar. Similarly, prepending "client_" to "secret" so its clear we're returning a client secret on creation.

Lmk what you think.

proto/viam/app/v1/app.proto
maxhorowitz: I am pretty heavily leaning towards the ability to delete along with adding a double confirm step in the CLI that makes it super clear to the user that its a dangerous operation ... the change was agreed upon in scope too and (I think) I made it clear that it was a new command (i.e. not from your existing work)

jr22 (approved):

PR #602, opened 2024-12-09
APP-7159: Add markdown_documentation field for inline module docs

michaellee1019:
The field addition is straightforward. I'm thinking about reducing the payload of APIs by separating out Model and ModelMarkdown into two different messages that APIs use. For example only GetModule, GetRegistryItem, and UpdateModule have a markdown field. But other APIs don't have it, especially List APIs. Thoughts?


michaellee1019:
> separating out Model and ModelMarkdown into two different messages that APIs use

Yeah I'm going to do this. Hold off on reviewing for now.

michaellee1019:
Ok this is ready for review now. I opted for an optional field on APIs for whether to include markdown or not. Updated the scope doc too. This way we have a public way to retrieve markdown documentation, and by default it is not returned in any APIs to keep response sizes low.

Abalaxy (approved):

proto/viam/app/v1/app.proto
```diff
@@ -1243,11 +1251,31 @@ message UpdateModuleResponse {
   string url = 1;
 }
 
+message ModuleUpdateMetadata {
+  // A list of models that are available in the module
+  repeated Model models = 1;
+  // The executable to run to start the module program
+  string entrypoint = 3;
```
ehhong, line 1258: 2?

proto/viam/app/v1/app.proto
```diff
@@ -1132,6 +1132,7 @@ message RegistryItem {
 
 message GetRegistryItemRequest {
   string item_id = 1;
+  optional bool include_markdown_documentation = 2;
```
ehhong, line 1135: when do we not want to include markdown documentation?

proto/viam/app/v1/app.proto
```diff
@@ -1243,11 +1251,31 @@ message UpdateModuleResponse {
...
+  bool draft = 3;
+}
+
 message Model {
   // The colon-delimited-triplet of the api implemented by the model
   string api = 1;
   // The colon-delimited-triplet of the model
   string model = 2;
+  // The markdown content describing the usage of the model
+  optional string markdown_documentation = 3;
```
ehhong, line 1278: would it be annoying to make a further nested object `ModelMetadata`? i can imagine we might want to include more things here later, and it feels a little jammed in `Model`

proto/viam/app/v1/app.proto
```diff
@@ -1155,6 +1156,12 @@ message UpdateRegistryItemRequest {
   string description = 3;
   Visibility visibility = 4;
   optional string url = 5;
+
+  oneof metadata {
+    ModuleUpdateMetadata module_update_metadata = 6;
+    MLModelUpdateMetadata ml_model_update_metadata = 7;
+    MLTrainingUpdateMetadata ml_training_update_metadata = 8;
```
ehhong, line 1163: why are these called __UpdateMetadata?

proto/viam/app/v1/app.proto
```diff
@@ -1280,6 +1308,7 @@ message UploadModuleFileResponse {
 message GetModuleRequest {
   // The id of the module (formatted as prefix:name where prefix is the module owner's orgid or namespace)
   string module_id = 1;
+  optional bool include_markdown_documentation = 2;
```
ehhong, line 1311: [just curious] do we have any plans to eventually deprecate `GetModule`?

proto/viam/app/v1/app.proto
ehhong: also, what was the decision to have separate documentation for each model rather than the entire module?

proto/viam/app/v1/app.proto
michaellee1019: Actually for fleet we are never going to use this in the current scope. But it will be available if we decide to show markdown on the registry pages at some point. It also makes it available to external API callers if needed.I made it so that by default, markdown is not returned so that API payloads are smaller.

My plan is to return markdown in the internal builtins service (using the same technical approach as [Fancy Config Tech Spec](https://docs.google.com/document/d/1PMRix3pT0Wu7NE5A0NCCZ9YhQVaJY9p0-oKxKIAsMeI/edit?tab=t.0#heading=h.pkgijd3tsb7t)

The alternative of the bool is to make separate messages for Request messages and Response messages. I tried that in previous commits in this PR. Its really gross and has backwards compatibility issues.

proto/viam/app/v1/app.proto
michaellee1019: Explained in this [scope ammendment](https://docs.google.com/document/d/1XPXGwMfxzMkt7tFJeU6ibg1fBh7vl9UvuQokiq6Oiq8/edit?tab=t.0#bookmark=id.o8k9d7813hlq)

```
I made a mistake in the scope and used the same proto messages for Update and Get APIs. Some fields are not allowed to be updated so I have created different messages to separate get/update
```

proto/viam/app/v1/app.proto
michaellee1019: Oops thank you!

proto/viam/app/v1/app.proto
michaellee1019: I think its fine to have high level fields for now. There will be an attributes field added in the coming months but still will be 1-to-1 to a model. I think when nesting protos, it makes sense when you need a list, not much rationale otherwise.

> documentation for each model rather than the entire module

These are actually two projects that will both be done. Model-level documentation is higher priority because it solves the problem of users configuring machines. Later we will support markdown at the module level so that the module detail page can be more content rich than just the `description` field we have now which only allows unformatted text.

Inline Model Documentation Epic (this PR): https://viam.atlassian.net/browse/APP-6907
Embed Markdown in registry items Epic (future project): https://viam.atlassian.net/browse/APP-2349

proto/viam/app/v1/app.proto
michaellee1019: No official plan but I do want to migrate the CLI to use the registry item APIs. Depends on effort. I was going to do it in the fancy config project. I'm going to pickup the CLI changes for this project next week and consider doing it then.

One of the blocking things was to add the `ModuleUpdateMetadata` message which is one reason why I included in this PR so that I have the option to do it.

michaellee1019: As soon as the CLI is off, the UI would need to migrate. And then probably give it another ~2-3mo for people to naturally upgrade to new versions of the CLI.

proto/viam/app/v1/app.proto
ehhong: ohh i see. i didn't understand which project this was for

proto/viam/app/v1/app.proto
ehhong: ah makes sense, so these are the editable fields per registry item. [opt, nit] would `UpdateModuleMetadata` or `ModuleMetadataUpdate` be clearer?

proto/viam/app/v1/app.proto
```diff
@@ -1341,6 +1370,7 @@ message Uploads {
 message ListModulesRequest {
   // The id of the organization to return private modules for.
   optional string organization_id = 1;
+  optional bool include_markdown_documentation = 2;
```
ehhong, line 1373: i don't think `ListModules` would ever want to return markdown, but does this have to do with this concern? 
> The alternative of the bool is to make separate messages for Request messages and Response messages. I tried that in previous commits in this PR. Its really gross and has backwards compatibility issues.

ehhong (approved):
thank you for all the clarifications! looks good

proto/viam/app/v1/app.proto
michaellee1019: Yeah this runs into the same issue that I was trying to fix by making a separate `Model` message. In order to not have a `markdown_documentation` field the response would have to be changed to a different message and cause backwards incompatibility. I could just remove this bool and always return a nil markdown. But I like the consistency of having the bool on all APIs.

proto/viam/app/v1/app.proto
michaellee1019: Yeah I can change to `UpdateModuleMetadata`. Makes sense

PR #604, opened 2024-12-10
Remove setup script

benjirewis:


Otterverse (approved):
LGTM, assuming the setup from RDK is sufficient and so the remaining makefile targets still work.

PR #607, opened 2024-12-17
APP-7232 Fix fragment usage protos

ehhong:
### Overview
Bozo move here. Fixing some mistakes that were not caught during review. These values are not being used anywhere yet so breaking change should be okay.


proto/viam/app/v1/app.proto
```diff
@@ -844,10 +844,10 @@ enum FragmentErrorType {
 }
 
 message FragmentUsage {
-  int32 fragment_id = 1;
-  int64 organizations = 2;
-  int64 machines = 3;
-  int64 machines_in_current_org = 4;
```
ehhong, line 850: @michaellee1019 do we have a standard on when we're using `int32` versus `int64`? seems pretty inconsistent right now. i realized that `int64` casts to a string in the frontend protos, which forces us to do a cast to use the values. i'm leaning towards using `int32` unless we're convinced we really need `int64`

michaellee1019: Huh I did not realize that. I've used int64 for other things and this never came up. I believe you and fine to move to int32. We have quite a while before 2.1 Billion machines use a fragment. I'm optimistic though. Max length is really only the reason.

michaellee1019 (approved):

proto/viam/app/v1/app.proto
mcous: FWIW `long_type_string` a buf configuration option that we could choose to disable. IIRC `int64` it will be a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) if we turn this off.

https://github.com/viamrobotics/app/blob/4af3eb4a1e3acd6c97346a63aca49a8b9fb92e4b/web-client/buf.gen.yaml#L5

> ```yaml
> version: v1
> plugins:
>   - name: ts
>     out: .
>     opt: generate_dependencies,force_optimize_code_size,long_type_string,ts_nocheck
>     strategy: all
> ```

mcous (approved):

proto/viam/app/v1/app.proto
ehhong: good to know, i think for this use case we should stick with `int32`, but good to know for the future

Abalaxy (approved):

proto/viam/app/v1/app.proto
michaellee1019: Also @mcous do you think at some point we could use the TS SDK directly? The SDK is already correctly using bigint for example: https://ts.viam.dev/classes/appApi.RegistryItem.html#totalExternalRobotUsage

PR #608, opened 2024-12-19
Data-3460: Adding GetInferenceResponse

etai-shuchatowitz:
Context: https://viam.atlassian.net/browse/DATA-3460?atlOrigin=eyJpIjoiMmVjNmJkOWU3MTlkNDA3ODgwODdiMWE4OTYxYWYwNjkiLCJwIjoiaiJ9

Added FlatTensors, Annotations and an Extra struct (so that we can add any potential metadata we want) to the GetInferenceResponse.


proto/viam/app/mlinference/v1/ml_inference.proto
```diff
@@ -20,4 +22,9 @@ message GetInferenceRequest {
   string organization_id = 4;
 }
 
-message GetInferenceResponse {}
+message GetInferenceResponse {
+  viam.service.mlmodel.v1.FlatTensors output_tensors = 1;
+  viam.app.data.v1.Annotations annotations = 2;
+  // anything else you want to say
+  google.protobuf.Struct extra = 99;
```
tahiyasalam, line 29: I think there's pretty low risk of needing to deprecate this field, but I would hate to introduce a backwards-incompatible breaking change over something as minor as an extra field. Is this intended to be used for something immediately? If not, I would err on just adding it as needed. WDYT?

proto/viam/app/mlinference/v1/ml_inference.proto
```diff
@@ -20,4 +22,9 @@ message GetInferenceRequest {
   string organization_id = 4;
 }
 
-message GetInferenceResponse {}
+message GetInferenceResponse {
+  viam.service.mlmodel.v1.FlatTensors output_tensors = 1;
```
tahiyasalam, line 26: Wooh, awesome. Thanks for maintaining parity with the ML model proto defs :)

tahiyasalam (commented):
Just one Q for you, otherwise looks good!

proto/viam/app/mlinference/v1/ml_inference.proto
etai-shuchatowitz: OK! That makes sense to me. Just thought it might be helpful because I find myself wanting to pass more info around in the MLModel in delphi. But we can add it if we need.

tahiyasalam: If we wanted to pass more info around in the MLModel in Delphi, wouldn't we then be working with the ML model proto? https://github.com/viamrobotics/api/blob/main/proto/viam/service/mlmodel/v1/mlmodel.proto

etai-shuchatowitz: Yeah, fixed it up! Thanks for this!

tahiyasalam (approved):

PR #610, opened 2024-12-30
APP-7267 APP-7269 Add version support to GetFragment

ehhong:
### Overview
in `GetFragment`:
- support returning versions
- support returning tags
- support returning usage for the requested version
- support requesting a version


proto/viam/app/v1/app.proto
```diff
@@ -870,11 +883,14 @@ message ListFragmentsResponse {
 message GetFragmentRequest {
   string id = 1;
   string current_organization_id = 2;
+  optional string version = 3;
 }
 
 message GetFragmentResponse {
   Fragment fragment = 1;
-  FragmentUsage fragment_usage = 2;
+  repeated FragmentUsage usages = 2;
```
ehhong, line 891: i think this is a breaking change, but i think it should be okay if it's resolved before the list machines using fragment epic is launched

michaellee1019: I think the reason we initially did a single item is that we could return a mapping of version usage in the FragmentUsage message. I don't think the [Fragment Version Tech Spec](https://docs.google.com/document/d/1K71WS1a1t9alIf3JJ3fHQr9Nq-ltqPmTTxz4G11vaFY/edit?tab=t.0) prescribes the exact message but we are planning on storing it as a map based on the [Fragment Usage Stats Tech Spec](https://docs.google.com/document/d/1zw_bQl4BdU7V6XitxfkrG6ML0b4FtQgBDNX45_PtLUg/edit?tab=t.0#heading=h.kbgutx5vgkgq).

proto/viam/app/v1/app.proto
```diff
@@ -870,11 +883,14 @@ message ListFragmentsResponse {
...
+  repeated FragmentVersion versions = 3;
+  repeated FragmentTag tags = 4;
```
michaellee1019, line 893: Should these be nested in the Fragment message? Or are you doing this to only return these fields on GetFragment?

proto/viam/app/v1/app.proto
```diff
@@ -822,6 +822,18 @@ message FragmentHistoryEntry {
   AuthenticatorInfo edited_by = 4 [(tagger.v1.tags) = "bson:\"edited_by\" json:\"edited_by\""];
 }
 
+message FragmentVersion {
+  string version = 1;
```
mcous, line 826: @michaellee1019 we talked about this in a dev jam before the holidays: you had some thoughts on aligning our language with registry modules. What terms do we use there, and do you have any thoughts about what we should use for fragments?

I remember you mentioning "version", "tag", and "revision" as different terms

mcous (commented):
LGTM, but would like ML to sound off so we're aligned with registry language

proto/viam/app/v1/app.proto
ehhong: oh hm, were you thinking something like this? for my understanding, is the primary advantage of this that it's closer to our DB storage schema?
```proto
message FragmentUsageStats {
  int32 organizations = 2;
  int32 machines = 3;
  int32 machines_in_current_org = 4;
}

message FragmentUsage {
  string fragment_id = 1;
  map<string, FragmentUsageStats> usage = 2;
}
```

proto/viam/app/v1/app.proto
ehhong: good q. looking at other uses of `Fragment`, if i nest it in `Fragment` the tags will also be returned with:
`GetFragmentHistory`
`ListFragments`
`CreateFragment`
`UpdateFragment`

do we think we'll want tag info for all of these endpoints? happy to nest if so

proto/viam/app/v1/app.proto
michaellee1019: I think there are a few options. Mainly we have to decide what the right balance between returning overall metrics along with version level metrics. I believe we still need to return overall metrics for a few reasons, mainly that the Fragment list page will show overall metrics:
1. The way you have it in the PR code now, all metrics are versioned and the only way to get overall metrics (usage across all versions) returned is to put in a key like "all" which isn't necessarily clean.
2. What you proposed above but then we have a similar issue with needing to store "all".
3. Could do this which has fields for revisions and overall:

```
message FragmentUsageStats {
  int32 organizations = 2;
  int32 machines = 3;
  int32 machines_in_current_org = 4;
}

message FragmentUsage {
  string fragment_id = 1;
  FragmentUsageStats fragment_usage = 2;
  map<string, FragmentUsageStats> fragment_revision_usage = 3;
}
```

LMK what you think. We could also change `fragment_usage` to `overall_usage` to be more clear but the way I wrote it out is not a breaking change.

proto/viam/app/v1/app.proto
michaellee1019: Great point, thanks for reminding me. I think here we'd keep the terms "version" and "tag". Whenever we want to combine the two into one field, Such as A) in the robot part config or B) in the metrics storage and protos, we should use the "revision" term. We can come up with a better name too, but something like "version_or_tag" should be avoided I think.

proto/viam/app/v1/app.proto
michaellee1019: Yeah I think for consistency lets put into the Fragment message. The size of the data transmitted is not going to increase that much. We might not use the fields on app.viam.com in some APIs but I think it will be useful for any API customers and prevent having to call Get after List, for example.

proto/viam/app/v1/app.proto
ehhong: ah, okay. i didn't realize that we also needed to show overall metrics. this makes sense!

proto/viam/app/v1/app.proto
ehhong: i thought version was the name for revision_or_tag?

proto/viam/app/v1/app.proto
ehhong: @michaellee1019 i updated to your suggestion. i renamed to `overall_usage` because i think it's a breaking change anyway to move the stats out of the original `FragmentUsage` message

proto/viam/app/v1/app.proto
ehhong: moved tags into fragment!

proto/viam/app/v1/app.proto
michaellee1019: Oops. You are right.
Revision: The fragment is saved and given an auto-incremented revision number
Tag: A string name that is a pointer to a Revision.
Version: A reference to either a Revision number or Tag string.

proto/viam/app/v1/app.proto
```diff
@@ -843,11 +856,16 @@ enum FragmentErrorType {
...
+  int32 organizations = 1;
+  int32 machines = 2;
+  int32 machines_in_current_org = 3;
+}
+
 message FragmentUsage {
   string fragment_id = 1;
-  int32 organizations = 2;
-  int32 machines = 3;
-  int32 machines_in_current_org = 4;
```
michaellee1019, line 850: I forgot this was breaking. But I think we can handle that. I think its better than keeping these fields AND having a separate object for versions

proto/viam/app/v1/app.proto
michaellee1019: TLDR: I think the following protos use the right terminology. The rest of the changes can stay as is because the version field is either referencing a revision or a tag:

```
message FragmentRevision {
  string revision = 1;
  google.protobuf.Timestamp created_at = 2;
}

message FragmentTag {
  string tag = 1;
  string revision = 2;
  google.protobuf.Timestamp created_at = 3;
  google.protobuf.Timestamp updated_at = 4;
}
```

ehhong: got it, updated!

proto/viam/app/v1/app.proto
```diff
@@ -870,11 +888,13 @@ message ListFragmentsResponse {
 message GetFragmentRequest {
   string id = 1;
   string current_organization_id = 2;
+  optional string version = 3;
 }
 
 message GetFragmentResponse {
   Fragment fragment = 1;
   FragmentUsage fragment_usage = 2;
+  repeated FragmentRevision revisions = 3;
```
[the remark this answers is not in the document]
ehhong: no longer relevant

proto/viam/app/v1/app.proto
```diff
@@ -843,11 +859,13 @@ enum FragmentErrorType {
   FRAGMENT_ERROR_TYPE_CYCLE_DETECTED = 4;
 }
 
+// Cached fragment usage statistics
 message FragmentUsage {
   string fragment_id = 1;
   int32 organizations = 2;
   int32 machines = 3;
   int32 machines_in_current_org = 4;
+  string version = 5;
```
ehhong, line 868: optional?

proto/viam/app/v1/app.proto
```diff
@@ -843,11 +859,13 @@ enum FragmentErrorType {
   FRAGMENT_ERROR_TYPE_CYCLE_DETECTED = 4;
 }
 
+// Cached fragment usage statistics
```
anjinai, line 862: Do we want to add a comment here about how this will either return tag usages or revision usages?

proto/viam/app/v1/app.proto
```diff
@@ -822,6 +826,18 @@ message FragmentHistoryEntry {
...
+message FragmentRevision {
+  string revision = 1;
+  google.protobuf.Timestamp created_at = 2;
+}
+
+message FragmentTag {
+  string tag = 1;
+  string revision = 2;
+  google.protobuf.Timestamp created_at = 3;
+  google.protobuf.Timestamp updated_at = 4;
```
mcous, line 838: UI is not using these fields in latest design; given there's some quirks around trying to store this data, let's ditch it from the response

proto/viam/app/v1/app.proto
mcous: I like `optional` for intent a little more

michaellee1019: +1 to optional.

michaellee1019 (approved):

anjinai (approved):

proto/viam/app/v1/app.proto
ehhong: updated!

Abalaxy (approved):

PR #612, opened 2025-01-02
APP-6581: Support granular billing SKUs in protos

jr22:
APP-6581

Added granular values to enum and should match list in app: https://github.com/viamrobotics/app/blob/909d44bb2533b90dba5cd832a05f5476e4ad4de3/data/billing_models.go#L38-L87

Also cross referenced designs: https://www.figma.com/design/Awj7KbWsDEQcjSPgMDWZV0/Billing?node-id=4995-1304&t=SgJMQlvYJiQLJDtk-0

Marked the less granular values as deprecated. Will bump app alongside this so I handle the deprecation warnings


ohEmily (approved):
LGTM! Thanks for taking care to deprecate the old fields.

PR #619, opened 2025-01-23
APP-7497: Add Button and Switch APIs

ethanlookpotts:
This adds the APIs for two new components: Button and Switch. Request from Eliot.

Next:
1. Add fake components to RDK for each
2. Add control cards to App for each

## Review requests

- Do you like the API? Anything about it you would change?


proto/viam/component/switch/v1/switch.proto
```diff
@@ -0,0 +1,45 @@
...
+
+  // Turns off the switch
+  rpc Off(OffRequest) returns (OffResponse) {
+    option (common.v1.safety_heartbeat_monitored) = true;
+    option (google.api.http) = {put: "/viam/api/v1/component/switch/{name}/off"};
+  }
+
+  // DoCommand sends/receives arbitrary commands
+  rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
+    option (google.api.http) = {post: "/viam/api/v1/component/gripper/{name}/do_command"};
```
randhid, line 28:
```suggestion
    option (google.api.http) = {post: "/viam/api/v1/component/switch/{name}/do_command"};
```

proto/viam/component/button/v1/button.proto
```diff
@@ -0,0 +1,32 @@
...
+service ButtonService {
+  // Pushes a button
+  rpc Push(PushRequest) returns (PushResponse) {
+    option (common.v1.safety_heartbeat_monitored) = true;
+    option (google.api.http) = {put: "/viam/api/v1/component/button/{name}/push"};
+  }
+
+  // DoCommand sends/receives arbitrary commands
+  rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
+    option (google.api.http) = {post: "/viam/api/v1/component/gripper/{name}/do_command"};
```
randhid, line 22:
```suggestion
    option (google.api.http) = {post: "/viam/api/v1/component/button/{name}/do_command"};
```

proto/viam/component/switch/v1/switch.proto
```diff
@@ -0,0 +1,45 @@
...
+    option (common.v1.safety_heartbeat_monitored) = true;
+    option (google.api.http) = {put: "/viam/api/v1/component/switch/{name}/on"};
+  }
+
+  // Turns off the switch
+  rpc Off(OffRequest) returns (OffResponse) {
+    option (common.v1.safety_heartbeat_monitored) = true;
+    option (google.api.http) = {put: "/viam/api/v1/component/switch/{name}/off"};
+  }
+
```
stevebriskin, line 25: Add a function to get the current state.

stevebriskin (changes requested):
one request. lgtm otherwise

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: Great catches, ty

proto/viam/component/switch/v1/switch.proto
```diff
@@ -0,0 +1,45 @@
...
+import "google/protobuf/struct.proto";
+
+option go_package = "go.viam.com/api/component/switch/v1";
+option java_package = "com.viam.component.switch.v1";
+
+// A SwitchService services switches associated with a machine
+service SwitchService {
+  // Turns on the switch
+  rpc On(OnRequest) returns (OnResponse) {
+    option (common.v1.safety_heartbeat_monitored) = true;
```
randhid, line 16: why is this heartbeat monitored? I think we're pushing to only do this for actuators.

randhid: It seems like this is analogous to board's SetGPIO API, which isn't session monitored, and, actually, DoCommand itself.  will it break anything if we want to remove it in the future? If so, let's just remove it now.

proto/viam/component/switch/v1/switch.proto
```diff
@@ -0,0 +1,45 @@
...
+
+import "common/v1/common.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/struct.proto";
+
+option go_package = "go.viam.com/api/component/switch/v1";
+option java_package = "com.viam.component.switch.v1";
+
+// A SwitchService services switches associated with a machine
+service SwitchService {
```
stevebriskin, line 13: Scope Creep:
Looking at this PR...I think we should model the API to support an N-state switch. So that 3-way switches and selector knobs can be supported. Both are very common. 
Suggesting an API update to something like:
void SetPosition(int)
int GetPosition()

Implementers can then configure a model with any number of states.
(offlined this with eliot as well)

stevebriskin (changes requested):

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: I'm happy to remove this for now - I had the same thought when copy-pasting gripper. It's easy enough to add in the future and there's not yet clear value.

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: I like this - please up the estimate to 6 engineer-weeks. 😈 

I will also add a method for getting the number of positions, so clients can render an interface or know the available options. This could also be `GetProperties`, but I like being explicit.
int GetNumberOfPositions()

proto/viam/component/button/v1/button.proto
ethanlookpotts: [2390a9e](https://github.com/viamrobotics/api/pull/619/commits/2390a9e8883a6a7c1e1d94e7d6b05000fcb0b5cd)

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: [2390a9e](https://github.com/viamrobotics/api/pull/619/commits/2390a9e8883a6a7c1e1d94e7d6b05000fcb0b5cd)

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: [f9fa7e6](https://github.com/viamrobotics/api/pull/619/commits/f9fa7e68d5ed633f2d0cbde243d669afe294d556)

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: [d2c7045](https://github.com/viamrobotics/api/pull/619/commits/d2c7045f187b6d08cb2b9b76a295a9643a2a611d)

proto/viam/component/switch/v1/switch.proto
ethanlookpotts: [d2c7045](https://github.com/viamrobotics/api/pull/619/commits/d2c7045f187b6d08cb2b9b76a295a9643a2a611d)

proto/viam/component/switch/v1/switch.proto
```diff
@@ -27,17 +34,29 @@ service SwitchService {
   }
 }
 
-message OnRequest {
+message SetPositionRequest {
   string name = 1;
+  uint32 position = 2;
```
randhid, line 39: hmm, do we want to allow negatives too?

ethanlookpotts: I don't see a reason to - position should just be a non-negative index representing the possible positions. The only case I can think of is if we wanted a negative position to be an error state, but I don't see that being useful. In error cases, the handlers should return errors.

proto/viam/component/switch/v1/switch.proto
```diff
@@ -0,0 +1,61 @@
...
+service SwitchService {
+  // Set the position of the switch
+  rpc SetPosition(SetPositionRequest) returns (SetPositionResponse) {
+    option (google.api.http) = {put: "/viam/api/v1/component/switch/{name}/set_position"};
+  }
+
+  // Get the position of the switch
+  rpc GetPosition(GetPositionRequest) returns (GetPositionResponse) {
+    option (google.api.http) = {put: "/viam/api/v1/component/switch/{name}/get_position"};
+  }
```
randhid, line 24: I don't love these API names, since it's more ambiguous than `ToggleTo` and `ToggleActive` but I don't care about this API enough to argue this, and I hope fleet enjoys owning their two new component APIs 😈

ethanlookpotts: Like we chatted about, I'm happy to change the names but don't feel strongly. Will leave as-is.

randhid: @stevebriskin we'll defer to you here.

proto/viam/component/switch/v1/switch.proto
randhid: I'm thinking of the most common toggle switch to me - the power drill direction - where that switch makes sense to me as a -1, 0, 1 switch. But  meh 🤷🏼‍♀️

stevebriskin (approved):

proto/viam/component/switch/v1/switch.proto
stevebriskin: IMO, all names are good enough and not worth debating in the interest of speed.

PR #622, opened 2025-01-24
APP-7502: Update FragmentHistoryEntry for fragment versioning

michaellee1019:
Adds the current config of the fragment history entry to the `GetFragmentHistory` API. This is a simplified version of an old PR that I found: https://github.com/viamrobotics/api/pull/578. As Emily P mentioned in that PR, trying to do all changes at once is going to cause a headache in app as well as SDKs. Lets deprecate and rename the existing fields later and get the main parts of the project done first.


Abalaxy (approved):
LGTM

proto/viam/app/v1/app.proto
```diff
@@ -820,6 +820,8 @@ message FragmentHistoryEntry {
   google.protobuf.Timestamp edited_on = 2 [(tagger.v1.tags) = "bson:\"edited_on\" json:\"edited_on\""];
   Fragment old = 3 [(tagger.v1.tags) = "bson:\"old\" json:\"old\""];
   AuthenticatorInfo edited_by = 4 [(tagger.v1.tags) = "bson:\"edited_by\" json:\"edited_by\""];
+  string revision = 5 [(tagger.v1.tags) = "bson:\"revision\" json:\"revision\""];
+  google.protobuf.Struct config = 6 [(tagger.v1.tags) = "bson:\"fragment\" json:\"fragment\""];
```
ehhong, line 824: double checking, this field `config` will eventually replace `old`? also, should the bson tags be `config` as well?

michaellee1019: Oops my bad about the tags, will update. Yes I think `old` will be deprecated. It doesn't make sense to drive the fragment version history off of `old` if we are storing that version's `config` as well. Plus its confusing to have the entire fragment in the response.

We have a lot of tasks to complete to migrate completely off of `old` and then we can deprecate it everywhere.

PR #629, opened 2025-02-11
APP-7615 Update list machine fragments to support versions

ehhong:
[tech spec
](https://docs.google.com/document/d/1K71WS1a1t9alIf3JJ3fHQr9Nq-ltqPmTTxz4G11vaFY/edit?tab=t.0#heading=h.nstno4i5n36z)

### Open Qs
- should we eventually deprecate `repeated string additional_fragment_ids = 2;`? or should we support both fields?
- which struct name?
```proto
message AdditionalFragment {
  string fragment_id = 1;
  string version = 2;
}

// or

message FragmentRequest {
  string fragment_id = 1;
  string version = 2;
}
```


proto/viam/app/v1/app.proto
```diff
@@ -976,13 +977,19 @@ message ListRobotsRequest {
   string location_id = 1;
 }
 
+message FragmentRequest {
```
mcous, line 980: **quibble:** I like your `AdditionalFragment` idea for the interface/field name symmetry

proto/viam/app/v1/app.proto
```diff
@@ -976,13 +977,19 @@ message ListRobotsRequest {
...
+}
+
 message ListMachineFragmentsRequest {
   // the machine_id used to filter fragments defined in a machine's parts.
   // Also returns any fragments nested within the fragments defined in parts.
   string machine_id = 1;
   // additional fragment_ids to append to the response. useful when needing to view fragments that will be
   // provisionally added to the machine alongside existing fragments.
   repeated string additional_fragment_ids = 2;
+  repeated FragmentRequest additional_fragments = 3;
```
mcous, line 992: **agreement:** yup, I think adding the field is the right move. As discussed with ML, let's hold off on any sort of `deprecated` or `reserved` keyword for the old field - we can do the least to make sure the old field doesn't break until we can ensure its usage is removed from the app

mcous (approved):

proto/viam/app/v1/app.proto
michaellee1019: Do you mean AdditionalFragment would be the name of this message? My thought here is that I don't like the word Request because it implies that its a message used at the top level of the API

ehhong: yeah, good point. i'm going to rename this to `AdditionalFragment`

proto/viam/app/v1/app.proto
```diff
@@ -1046,13 +1047,20 @@ message ListRobotsRequest {
   string location_id = 1;
 }
 
+message AdditionalFragment {
+  string fragment_id = 1;
+  optional string version = 2;
```
ehhong, line 1052: i made this optional. does that seem right to everyone?

Abalaxy (approved):
Just a couple of naming questions, looks good to merge once those are addressed/discussed!

PR #633, opened 2025-02-24
APP-7667: Adding org name to the Locations struct

RoxyFarhad:
- we need this to be in the public API as that is where the locations struct is defined


jr22 (approved):
this struct is actually only used in a few apis (getLocations, createLocation, listLocation) so this feels fine to me. thanks!

PR #634, opened 2025-02-25
RSDK-9861 Add tunnel ports and connection timeouts to network config

benjirewis:


proto/viam/app/v1/robot.proto
```diff
@@ -139,12 +139,18 @@ message NetworkConfig {
...
+  repeated TrafficTunnelEndpoint traffic_tunnel_endpoints = 6;
 }
 
 message SessionsConfig {
   google.protobuf.Duration heartbeat_window = 1;
 }
 
+message TrafficTunnelEndpoint {
+  int32 port = 1;
+  int32 connnection_timeout_seconds = 2;
```
benjirewis, line 151: Part of me wonders if this should be a `string` and be a duration in the same way as `heartbeat_window`. You didn't scope this right, @cheukt ? i.e. we didn't agree on a specific API?

cheukt: ya a `google.protobuf.Duration` makes sense

benjirewis: Cool; switched and renamed `connection_timeout_seconds` -> `connection_timeout`.

benjirewis:
Gonna test with RDK PR before I merge.

PR #636, opened 2025-02-28
RSDK-9862 Add ListTunnels to robot API

benjirewis:


proto/viam/robot/v1/robot.proto
```diff
@@ -135,6 +140,17 @@ message TunnelResponse {
   bytes data = 1;
 }
 
+message ListTunnelsRequest {}
+
+message ListTunnelsResponse {
+  repeated Tunnel tunnels = 1;
+}
+
+message Tunnel {
```
benjirewis, line 149: `Tunnel` is an exact duplicate of [`TrafficTunnelEndpoint`](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/robot.proto#L149-L152) from the app proto. I should have put `TrafficTunnelEndpoint` in the common proto instead of in the app proto (I don't _think_ I can import from app proto here without weird issues,) but I was dumb and lacked foresight. I can move `TrafficTunnelEndpoint` to common without much trouble, but it will be a breaking change that will require admin approval (Steve, I'm assuming.) Let me know what you think.

cheukt: I think it is ok to have it be a copy - they might diverge at some point, and do not have to match

benjirewis: A good point there that it actually might be good to _not_ have `Tunnel` and `TrafficTunnelEndpoint` not be identical. We may want to be able to configure tunnels with different fields than the fields we use to represent their configuration to users or something.

PR #648, opened 2025-03-13
RSDK-10250 add mesh type to geometry oneof

raybjork:
This work adds the Mesh type to the Geometry oneof


nfranczak (approved):

nfranczak:
Not sure why this failed. Might need the `protos-compiled` tag?

PR #661, opened 2025-03-31
DATA-3781: Deprecate AddTagsToBinaryDataByFilter, RemoveTagsFromBinaryDataByFilter

vpandiarajan20:
Marks Add/RemoveTagsToBinaryDataByFilter as deprecated. I'm actually not sure if I should be doing anything else to deprecate these endpoints, so please let me know!


vijayvuyyuru (approved):
These changes look fine to me. Not sure what the timeline is on fully removing these. Theres a reasonable chance youll need to update RDK and app to have no lints around these methods on the newest API version, otherwise itll block other peoples merges. Im not as sure about the SDKs outside of Go. Also not sure if this needs to be posted in breaking changes, I assume these endpoints arent really used?

tahiyasalam (changes requested):
requesting changes so we can ensure we go through the punch list before deprecating! ty!!

PR #663, opened 2025-04-01
APP-7929 add markdown description to moduleversion

gmulz:
Adds `markdown_description` to `ModuleVersion`

Sidenote: Can someone please tell me why VersionHistory and ModuleVersion are two separate protos even though they are virtually the same thing?


Abalaxy (approved):
No idea on the different protos, maybe they were different in the past?

PR #667, opened 2025-04-04
APP-7877 Add fragment fragments endpoint

ehhong:


DTCurrie (approved):
Did we bump `connect-es` or something? Looks like a bunch of unrelated changes in the `gen` code. LGTM otherwise. Maybe the `rpc` could be called `ListNestedFragments` for brevity? But I am assuming that was already considered/discussed.

ehhong:
> Did we bump `connect-es` or something? Looks like a bunch of unrelated changes in the `gen` code. LGTM otherwise. Maybe the `rpc` could be called `ListNestedFragments` for brevity? But I am assuming that was already considered/discussed.

@DTCurrie will regenerate the code. name was not discussed and i don't love the current name so will use `ListNestedFragments` -- thanks!

PR #668, opened 2025-04-07
APP-7932: Add rename module protos

Abalaxy:
[Ticket](https://viam.atlassian.net/browse/APP-7932)

Add protos for the rename module endpoint. `TransferRegistryItem` was used as a model, since the `app` repo endpoint should be very similar to it


proto/viam/app/v1/app.proto
```diff
@@ -1434,6 +1436,13 @@ message CreateModuleResponse {
   string url = 2;
 }
 
+message RenameModuleRequest {
```
gmulz, line 1439: the scope has this specific to Registry Items, not modules. seems like we would want this to apply to ML models as well so I think we should rename 

image

proto/viam/app/v1/app.proto
```diff
@@ -1434,6 +1436,13 @@ message CreateModuleResponse {
...
+  string module_id = 1;
+  string new_module_name = 2;
+}
+
+message RenameModuleResponse {}
```
mcous, line 1444: **suggestion:** should this return the updated module? It would allow us to optimistically update the query cache on the frontend for a more responsive UX during a rename

mcous (approved):
LGTM! One small request: I think it'd be cool if we could return the updated resource in the response

proto/viam/app/v1/app.proto
Abalaxy: Interesting, the scope also specifically calls out that ML models are a non-goal. I see that we don't allow transferring ownership of ML models either, is that just a "we haven't implemented that yet" thing, or a "there are differences between ML models and modules that makes the implementation non-trivially different" thing @michaellee1019? And is this API naming convention in the config related to eventually supporting ML models, or around removing the sub-concept of "modules" and folding them up into registry items?

proto/viam/app/v1/app.proto
gmulz: +1 to this, was thinking a similar thing. returning the updated object is somewhat standard in CRUD operations

proto/viam/app/v1/app.proto
michaellee1019: From an API perspective, lets name it "registry" and be able to support any types that are in the registry. On the backend and frontend, we can not implement it and wait for later (throw an exception that the "ml_model" and "training_script" types are not supported. This way we don't have to make a breaking API change when we do want to support transfer and rename for the other types.

I have a future project to address the feature discrepancies called "Universal Version Level Detail Page" which will force us to be at feature parity between all types. So it will be needed in the next ~6mo.

proto/viam/app/v1/app.proto
michaellee1019: Fine with this change.

proto/viam/app/v1/app.proto
gmulz: i think the non-goals in the scope are more "not yet" rather than explicitly "not supporting this"

proto/viam/app/v1/app.proto
Abalaxy: The interesting thing is, in the ownership transfer code (which is the blueprint for this feature), the call doesn't actually carry out the transfer, it just adds it to a queue and the transfer happens in a scheduled job every 5 seconds. I'm assuming there was a concern about processing time? Looks like it hits something called a package server, so that might be the cause. Does a similar processing time concern apply here as well, or is a direct re-name fine? @michaellee1019

michaellee1019: Yeah, transfer is an organization owner change, so it required a background job flow to move files from one org to the other. For rename, that isn't needed.

proto/viam/app/v1/app.proto
```diff
@@ -1425,6 +1427,15 @@ message DeleteRegistryItemRequest {
 
 message DeleteRegistryItemResponse {}
 
+message RenameRegistryItemRequest {
+  string module_id = 1;
+  string new_module_name = 2;
```
michaellee1019, line 1432:
```suggestion
  string item_id = 1;
  string new_name = 2;
```

to match scope document

gmulz (approved):

PR #673, opened 2025-04-09
DATA-3916 Remove reason

vijayvuyyuru:
The reason column currently includes internal error details. If and when we introduce serving failure details in the API, that should be something different than what we currently have stored.


n0nick (approved):
LGTM and thank you.

I suggest adding to the PR description, for future reference:

> The reason column currently includes internal error details. If and when we introduce serving failure details in the API, that should be something different than what we currently can.

PR #677, opened 2025-04-14
nick-small-camera-api-change

nicksanford:


proto/viam/component/camera/v1/camera.proto
```diff
@@ -69,6 +72,10 @@ message GetImageResponse {
 message GetImagesRequest {
   // Name of a camera
   string name = 1;
+  repeated string desired_source_names = 2;
+  repeated Format desired_formats = 3;
```
nicksanford, line 76: don't do this

proto/viam/component/camera/v1/camera.proto
```diff
@@ -69,6 +72,10 @@ message GetImageResponse {
 message GetImagesRequest {
```
nicksanford, line 72: Will have far-reaching implications accross UIs.

PR #679, opened 2025-04-15
Consolidate camera api demo

nicksanford:


proto/viam/component/camera/v1/camera.proto
```diff
@@ -12,29 +12,47 @@ option java_package = "com.viam.component.camera.v1";
...
   }
 
-  // GetPointCloud returns a point cloud from a camera of the underlying robot. A specific MIME type
-  // can be requested but may not necessarily be the same one returned.
+  // GetPointCloud returns a point cloud from a camera of the underlying robot.
+  // A specific MIME type can be requested but may not necessarily be the same
+  // one returned.
   rpc GetPointCloud(GetPointCloudRequest) returns (GetPointCloudResponse) {
+    option deprecated = true;
     option (google.api.http) = {get: "/viam/api/v1/component/camera/{name}/point_cloud"};
```
nicksanford, line 41: We wouldn't deprecate these APIs immediately as part of this proposal, but rather, we would remove them at the end of the project once all important camera models have been migrated to the new API that can express all of them.

proto/viam/component/camera/v1/camera.proto
```diff
@@ -12,29 +12,47 @@ option java_package = "com.viam.component.camera.v1";
...
     option (google.api.http) = {get: "/viam/api/v1/component/camera/{name}/images"};
   }
 
-  // RenderFrame renders a frame from a camera of the underlying robot to an HTTP response. A specific MIME type
-  // can be requested but may not necessarily be the same one returned.
+  // RenderFrame renders a frame from a camera of the underlying robot to an
+  // HTTP response. A specific MIME type can be requested but may not
+  // necessarily be the same one returned.
   rpc RenderFrame(RenderFrameRequest) returns (google.api.HttpBody) {
+    option deprecated = true;
```
seanavery, line 32: Wondering why we needed this custom API endpoint for the frontend to begin with since the web app can handle GRPC responses.

PR #690, opened 2025-05-02
APP-8187: Add apps to Module API

jr22:
Updating the [API Module struct](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/app.proto#L1589-L1619) to include apps info similar to the [UpdateModuleRequest](https://github.com/viamrobotics/api/blob/29c7a2e5ec55ec1bb82b4ddae40d66396490cbd8/proto/viam/app/v1/app.proto#L1501).


michaellee1019 (approved):

bashar-515 (approved):

proto/viam/app/v1/app.proto
```diff
@@ -1292,6 +1292,8 @@ message ModuleVersion {
   optional string first_run = 5;
   // The markdown documentation for this version of the module
   optional string markdown_description = 6;
+  // A list of applications associated with the module
+  repeated App apps = 7;
```
jr22, line 1296: @michaellee1019 rerequesting you since i added this field in a few new places since you approved (see [this commit](https://github.com/viamrobotics/api/pull/690/commits/9f9f8a8a227610f94d2f9ff276342f078dd98f43)). 

open q-- do you think it makes sense for this new field to be `optional`? when i was testing it in app, there were a few places in the ts that didnt populate with the new field in the request and caused the gRPC request to fail. i wonder if thats desired behavior for this? should we make it optional instead?

jr22: [this](https://github.com/viamrobotics/app/pull/8392/files#diff-889d8ba0733d6d39be33ec6169a06faa1a40dd10e59ba1e7cd7967e9a740cf0fR183) was the line i needed to add to unbork updating modules in the UI

michaellee1019: Im pretty sure you cannot have a `repeated optional`, only repeated or optional. So its fine as is. Yeah the TS linter is pretty annoying in that you are required to place each field into several places once a new field is added in this API.

michaellee1019 (approved):

proto/viam/app/v1/app.proto
jr22: ah ok! thanks michael

PR #692, opened 2025-05-13
[APP-8225] Change UpdateRobot API Name and Location to optional

JosephBorodach:
# Description
Per the Scope API section, update the UpdateRobotRequest proto such that name and location are both optional. We will enforce that at least one must be included in code

**See ticket [here](https://viam.atlassian.net/browse/APP-8225)**


JosephBorodach:
Not doing this ticket for now. tl;dr it's a nice to have but not worth breaking changes. See the ticket for the longer explanation

PR #695, opened 2025-05-20
APP-8003 Add world object store service

DTCurrie:
Introducing a new service type, the `WorldObjectStoreService`.

This service will be used by visualization modules to return information about the machine's world object store to be rendered client-side. 

- `ListObjects` returns a list of `bytes uuid` representing a list of the object ID's in the machine's world store
  - The machine will handle adding/removing entries from the returned array of `uuids`, the client will be expected to manage adding and removing `WorldObjects` from the renderer.
- `GetObject` returns `WorldObject` information by `uuid` in order to be rendered

### DIFF
--- a/proto/viam/service/worldobjectstore/v1/world_object_store.proto
+++ b/proto/viam/service/worldobjectstore/v1/world_object_store.proto
@@ -1,0 +1,62 @@
+syntax = "proto3";
+
+package viam.service.worldobjectstore.v1;
+
+import "common/v1/common.proto";
+import "google/protobuf/struct.proto";
+
+option go_package = "go.viam.com/api/service/worldobjectstore/v1";
+option java_package = "com.viam.service.worldobjectstore.v1";
+
+message WorldObjectPoints {
+    repeated float array = 1;
+}
+
+message WorldObjectLine {
+    repeated float array = 1;
+}
+
+message WorldObject {
+  bytes uuid = 1;
+  string name = 2;
+  common.v1.PoseInFrame pose = 3;
+
+  // Can hold information like color, opacity, points colors, collisionAllowed, etc
+  google.protobuf.Struct metadata = 4;
+
+  oneof geometry {
+    common.v1.RectangularPrism box = 5;
+    common.v1.Capsule capsule = 6;
+    common.v1.Sphere sphere = 7;
+    common.v1.Mesh mesh = 8;
+    WorldObjectPoints points = 9;
+    WorldObjectLine line = 10;
+  }
+}
+
+service WorldObjectStoreService {
+    // ListObjects returns all world object ids
+    rpc ListObjects(ListWorldObjectsRequest) returns (ListWorldObjectsResponse) {}
+
+    // GetObject returns a world object by id
+    rpc GetObject(GetWorldObjectRequest) returns (GetWorldObjectResponse) {}
+
+    // DoCommand sends/receives arbitrary commands
+    rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
+        option (google.api.http) = {post: "/viam/api/v1/service/worldobjectstore/{name}/do_command"};
+    }
+}
+
+message ListWorldObjectsRequest {}
+
+message ListWorldObjectsResponse {
+    repeated bytes uuids = 1;
+}
+
+message GetWorldObjectRequest {
+    bytes uuid = 1;
+}
+
+message GetWorldObjectResponse {
+    WorldObject object = 1;
+}


DTCurrie:
Removing some folks from review while I tweak things to avoid noise.

raybjork:
Was there a scope for this?  Would be helpful to understand the thinking behind it

DTCurrie:
> Was there a scope for this? Would be helpful to understand the thinking behind it

Yes sorry, meant to add a link in the description. Here you go: https://docs.google.com/document/d/1ionvyBa7x3HZU_rwDPBvrAUrQjBrP6sJ10Z_xbBTue8/edit?tab=t.0#heading=h.tcicyojyqi6c

The API design has been slightly tweaked based on testing @micheal-parks did, but for the most part things are the same.

raybjork (approved):
This looks good to me from a code perspective based on what was decided on in the scope doc.  I wonder though if it would be possible to put this under the app proto instead of the general service proto.  I expect this API to change frequently and it seems like there is a lot more leniency towards making breaking changes to the App API as compared to resources

DTCurrie:
> This looks good to me from a code perspective based on what was decided on in the scope doc. I wonder though if it would be possible to put this under the app proto instead of the general service proto. I expect this API to change frequently and it seems like there is a lot more leniency towards making breaking changes to the App API as compared to resources

Fair point, I brought this up to steve and he suggested it could live here because it is a new service resource type, but we can mark it as experimental for now, and allow some level of leniency until we have published some Viam-supported visualization modules and are confident with the API. 

@micheal-parks has also been testing visualizations with the assumption that this is how the API will work, so we have some real-world proof of concepts to work from. I ended up writing the code to implement the `fake` model for testing in the RDK. I can work with micheal to test these APIs using his current code and ensure they work as expected, then report back on the results.

I'd be happy to get some more opinions on this before moving forward, though, if you have anyone in mind.

proto/viam/service/worldobjectstore/v1/world_object_store.proto
```diff
@@ -0,0 +1,49 @@
...
+  string name = 1;
+
+  bytes uuid = 2;
+
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
+}
+
+message GetWorldObjectResponse {
+  common.v1.WorldState world_state = 1;
```
DTCurrie, line 48: @micheal-parks per Eliot's note I updated this, which is defined in the [common protos](https://github.com/viamrobotics/api/blob/main/proto/viam/common/v1/common.proto#L152)

micheal-parks:
There are a few minor problems - and one bigger problem - with using the existing `GeometriesInFrame` struct:

Minor issues:
* The `geometry` does not have a `bytes uuid` (can we add?)
* The `geometry` struct does not have a `metadata` object for drawing poses, colors, etc (can we add?)
* We have sphere, box, capsule, but would need `points` and `lines` to encompass what people draw
* Some things that people draw are not geometries (like poses represented as arrows)

The bigger issue is:
* ListWorldObjects was designed to return an array of uuids packed into bytes so that rapid inexpensive polling for store state could be done from a remote client
* The client decides which uuids are new and fetches geometry details only for added items instead of fetching complete details for all world objects on each poll
* An array of world states would be an array of arrays, and with this data structure it's easier to pack items into the geometries instead of the world states
* People often draw hundreds or thousands of objects with this tool - it's not tenable to fetch that much data per poll

We could just make this a geometry list store if the minor issues are solved, but IMO it's still weird because geometries are only some of the things people visualize

DTCurrie:
> There are a few minor problems - and one bigger problem - with using the existing `GeometriesInFrame` struct:
> 
> Minor issues:
> 
> * The `geometry` does not have a `bytes uuid` (can we add?)
> * The `geometry` struct does not have a `metadata` object for drawing poses, colors, etc (can we add?)
> * We have sphere, box, capsule, but would need `points` and `lines` to encompass what people draw
> * Some things that people draw are not geometries (like poses represented as arrows)
> 
> The bigger issue is:
> 
> * ListWorldObjects was designed to return an array of uuids packed into bytes so that rapid inexpensive polling for store state could be done from a remote client
> * The client decides which uuids are new and fetches geometry details only for added items instead of fetching complete details for all world objects on each poll
> * An array of world states would be an array of arrays, and with this data structure it's easier to pack items into the `GeometriesInFrame` instead of one object per world state which doesn't make much sense
> * This now makes it impossible to diff on the client
> * People often draw hundreds or thousands of objects with this tool - it's not tenable to fetch that much data per poll
> 
> We could just make this a geometry list store if the minor issues are solved, but IMO it's still weird because geometries are only some of the things people visualize

I noticed some of these issues and was thinking about how best to address them. We could flatten the world states into a single world state, then you have a single list of `geometries` and `transforms` to deal with. We would then have to add the fields you mentioned to `geometries`. I am a little hesitant to make alterations to that common proto to support this one service type, though. And that still has to introduce a new `WorldStates` type or something. 

I may revert the changes, then we can sync over email to decide on this quickly.

DTCurrie:
Updated per latest conversation:
 
> 1. New message type: `GeometryInFrame`, which is just one geometry instead of a repeated list
> 2. Add `uuid` and `metadata` to `PoseInFrame` and `GeometryInFrame`
> 3. Add `label` to `PoseInFrame`
> 4. Add `Line` and `Points` to `Geometry['geometryType']`. They have one field: array = Float32Array. This is essential for faster over-the-wire speeds, especially for point clouds
> 5. The world state store service is a list of `GeometryInFrame` / `PoseInFrame`

proto/viam/common/v1/common.proto
```diff
@@ -42,6 +42,8 @@ message Pose {
   double o_z = 6;
   // degrees
   double theta = 7;
+  // Label of the pose. If none supplied, will be an empty string.
+  string label = 8;
```
raybjork, line 46: this should be an optional

raybjork: but more importantly, why would we be adding this in the first place?

proto/viam/common/v1/common.proto
```diff
@@ -59,6 +61,10 @@ message Orientation {
 message PoseInFrame {
   string reference_frame = 1;
   Pose pose = 2;
+  // UUID of the geometry. If none supplied, will be an empty string.
+  bytes uuid = 3;
+  // Can hold information like color, opacity, points colors, collisionAllowed, etc...
+  google.protobuf.Struct metadata = 4;
```
raybjork, line 67: Same thing here.  I don't like adding fields to things that should just be geometry related.  Especially something like an unstructured metadata struct.  I would like to understand why these are necessary

proto/viam/common/v1/common.proto
```diff
@@ -113,6 +129,16 @@ message GeometriesInFrame {
   repeated Geometry geometries = 2;
 }
 
+// GeometryInFrame contains a geometry and the and the reference frame in which it was observed
+message GeometryInFrame {
```
raybjork, line 133: The purpose of the `GeometriesInFrame` object above is to allow for the singular case as well if one is provided, so this is almost entirely duplicative.  I again do not understand why a uuid and metadata is necessary

proto/viam/service/worldstatestore/v1/world_state_store.proto
```diff
@@ -0,0 +1,53 @@
...
+
+service WorldStateStoreService {
+  // ListWorldStates returns all world state object uuids
+  rpc ListWorldStateUUIDs(ListWorldStateUUIDsRequest) returns (ListWorldStateUUIDsResponse) {}
+
+  // GetWorldState returns a world state object by uuid
+  rpc GetWorldStateObject(GetWorldStateObjectRequest) returns (GetWorldStateObjectResponse) {}
+
+  // DoCommand sends/receives arbitrary commands
+  rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
```
raybjork, line 21: How are we ever supposed to be able to add things to this?

proto/viam/service/worldstatestore/v1/world_state_store.proto
```diff
@@ -0,0 +1,53 @@
...
+  // Name of the world object store service
+  string name = 1;
+
+  bytes uuid = 2;
+
+  // Additional arguments to the method
+  google.protobuf.Struct extra = 99;
+}
+
+message GetWorldStateObjectResponse {
```
raybjork, line 48: A couple things.  
- We already have a WorldState object, which this is named the same as but is different from.  This is confusing.  
- The metadata and UUID are entirely used within the context of this service, so no need to add them to the common structures, they can live in this scope

raybjork (changes requested):

proto/viam/service/worldstatestore/v1/world_state_store.proto
DTCurrie: I am confused, these were all already discussed in the email. Micheal is sick today so I guess we can discuss these further next week.

proto/viam/service/worldstatestore/v1/world_state_store.proto
DTCurrie: I am not sure what you are asking, but this how all `DoCommand`s are defined for components and services. Did I miss something? Or Is there something specific you think we should be adding for this service?

Examples:

```
  // DoCommand sends/receives arbitrary commands
  rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
    option (google.api.http) = {post: "/viam/api/v1/component/button/{name}/do_command"};
  }
 
 ...
 
   // DoCommand sends/receives arbitrary commands
  rpc DoCommand(common.v1.DoCommandRequest) returns (common.v1.DoCommandResponse) {
    option (google.api.http) = {post: "/viam/api/v1/service/navigation/{name}/do_command"};
  }
  ```

File: proto/viam/common/v1/common.proto (hunk 4 above)
Comment (DTCurrie) [reply proto/viam/common/v1/common.proto:133]: This was discussed in the email:

> There are a few minor problems - and one bigger problem - with using the existing GeometriesInFrame struct:
Minor issues:
> - The geometry does not have a bytes uuid (can we add?)
> - The geometry struct does not have a metadata object for drawing poses, colors, etc (can we add?)
> - We have sphere, box, capsule, but would need points and lines to encompass what people draw
> - Some things that people draw are not geometries (like poses represented as arrows)

>The bigger issue is:
> - ListWorldObjects was designed to return an array of uuids packed into bytes so that rapid inexpensive polling for store state could be done from a remote client
> - The client decides which uuids are new and fetches geometry details only for added items instead of fetching complete details for all world objects on each poll
> - An array of world states would be an array of arrays, and with this data structure it's easier to pack items into the GeometriesInFrame instead of one object per world state which doesn't make much sense
> - This now makes it impossible to diff on the client
> - People often draw hundreds or thousands of objects with this tool - it's not tenable to fetch that much data per poll

File: proto/viam/common/v1/common.proto (hunk 3 above)
Comment (DTCurrie) [reply proto/viam/common/v1/common.proto:67]: Per the email:

> Some things that people draw are not geometries (like poses represented as arrows)


We will want a way to provide rendering data like color and opacity for those cases.

File: proto/viam/service/worldstatestore/v1/world_state_store.proto (hunk 5 above)
Comment (DTCurrie) [reply proto/viam/service/worldstatestore/v1/world_state_store.proto:21]: That was in the scope:

>  // Add / remove objects de-scoped due to not being needed yet.
  // Existing examples above involve a module implementation of this service 
  // querying other resources and internally adding / removing objects.

File: proto/viam/service/worldstatestore/v1/world_state_store.proto (hunk 6 above)
Comment (raybjork) [reply proto/viam/service/worldstatestore/v1/world_state_store.proto:48]: Yeah lets discuss next week

File: proto/viam/common/v1/common.proto (hunk 2 above)
Comment (raybjork) [reply proto/viam/common/v1/common.proto:46]: Why can't we just use the existing Geometry type.  It has a `Pose` on it and a label.  The geometry_type can be undefined.

Comment (DTCurrie) [reply proto/viam/common/v1/common.proto:46]: Maybe @micheal-parks can answer more clearly, but my understanding is this is how the viking sanding project is rendering the pose arrows. I can't answer as to why that was preferred over using geometries, I am not sure if that was a viz team decision or not.

Comment (raybjork) [reply proto/viam/common/v1/common.proto:46]: Yes that is correct.  We did it for simplicity before, since none of this had to be in the proto and it was easier.  But it would be very simple to change out the internals to pass Geometry types instead.  

In general I'm pretty resistant to adding more than we need to to these structures, since complexity creep is something that the motion stack has suffered from in the past.

**Review by raybjork — APPROVED** [review]

**Comment by biotinker** [comment]
> Should we remove the Line and Points messages and instead leave them as multi/composite geometries?

While I'm not sold on needing a special `line` type and think it would be nice to implement multi geometries with lines as 0-radius capsules, `Points` we do need to make a change for. `Pointcloud` is notably absent from the Geometry message.

Currently PCDs are sent over the wire elsewhere as `bytes`. If we transmit points as `[]r3.Vector`, that omits potentially collision-relevant data from the pointcloud, specifically certainty values/cutoffs.

File: proto/viam/service/worldstatestore/v1/world_state_store.proto [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+  // ListUUIDs returns all world state transform UUIDs
+  rpc ListUUIDs(ListUUIDsRequest) returns (ListUUIDsResponse) {}
+
+  // GetTransform returns a world state transform by uuid
+  rpc GetTransform(GetTransformRequest) returns (GetTransformResponse) {}
+
+  // StreamTransformChanges streams changes to world state transforms
+  rpc StreamTransformChanges(StreamTransformChangesRequest) returns (stream StreamTransformChangesResponse) {
+    option (google.api.http) = {get: "/viam/api/v1/service/worldstatestore/{name}/stream_transform_changes"};
+  }
```
DTCurrie, line 23: Added this based on a conversation with @stevebriskin and @micheal-parks . The intention of this RPC is to allow visualization tools to open a stream and listen for change events from this service. We define a change event as one of the following:

```proto
// TransformChangeType is the type of change that has occurred for a transform.
enum TransformChangeType {
  TRANSFORM_CHANGE_TYPE_UNSPECIFIED = 0;
  TRANSFORM_CHANGE_TYPE_ADDED = 1;
  TRANSFORM_CHANGE_TYPE_REMOVED = 2;
  TRANSFORM_CHANGE_TYPE_UPDATED = 3;
}
```

When an implementor of this service is going to `Send` a change message, they must define what type of change event is occurring in the response:

```proto
message StreamTransformChangesResponse {
  TransformChangeType change_type = 1;
  common.v1.Transform transform = 2;

  // The field mask of the Transform that has changed, if any. For added transforms, this will be empty. For updated
  // transforms, this will be the fields that have changed. For removed transforms, this will be the Transform's UUID
  // path.  
  google.protobuf.FieldMask updated_fields = 3;
}
```

Note we use a `FieldMask`, which will allow the implementor only to include fields that were changed during a `TRANSFORM_CHANGE_TYPE_UPDATED` event, or just the `change_type` and `transform.uuid` fields for a `TRANSFORM_CHANGE_TYPE_REMOVED` event.

The field mask will enable the implementor to reduce the amount of data sent over the wire in a stream response to only the relevant data, allowing the client to understand how to manage each event it receives quickly. This should be particularly helpful when dealing with `Transform`s that have large amounts of geometry data (like point clouds) or have large quantities of geometries defined.

proto/viam/common/v1/common.proto
```diff
@@ -89,6 +89,14 @@ message Mesh {
   bytes mesh = 2;
 }
 
+message PointCloud {
+  bytes point_cloud = 1;
+}
+
+message Line {
+  repeated float segments = 1;
+}
```
DTCurrie, line 98: Do we still want to introduce the `Line` type if we are allowing `Transform` to have multiple `Geometry`s?

raybjork: I'm not sure what this has to do with Transform having multiple geometries

micheal-parks: should line be `bytes` too to match Pointcloud?

DTCurrie: Yeah, much easier to add it later than take it away.

PR #697, opened 2025-05-29
Allow metadata in Tunnel messages

cheukt:
will allow for sending metadata over the wire (things like sent time or debug flags), which will be super helpful in improving/debugging performance

but also open to adding the necessary fields directly (e.g. session_id, timestamp, debug). just thought doing it as a map is a bit easier and allows us to iterate quickly.

can't use grpc metadata for timestamp, since that is sent only once per streaming call.

### DIFF
--- a/proto/viam/robot/v1/robot.proto
+++ b/proto/viam/robot/v1/robot.proto
@@ -127,10 +127,12 @@ service RobotService {
 message TunnelRequest {
   uint32 destination_port = 1;
   bytes data = 2;
+  map<string, google.protobuf.Any>  metadata = 3;
 }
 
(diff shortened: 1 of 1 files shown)


ale7714 (approved):
👍

benjirewis (commented):
This LGTM, but sounds like we should gather more tunnel statistics with custom viam-servers and CLIs before merging this API? I think this will be good for the eventual prod version of collecting tunnel statistics.

dgottlieb:
What's the status of this @cheukt ? If we're not ready to merge now -- it seems we can close and re-open later? Seems easy enough to re-type if necessary.

PR #702, opened 2025-05-30
APP-8364: Add GetRobotPartByNameAndLocation API

jr22:
APP-8364

In APP-8364, we want to add support for a `/host/hostname` api for viam apps. In order to support this, we need a way to map a host -> machine ID so we can fetch the appropriate api keys. I thought about setting this via cookies once you select a machine etc but that would cause direct links to the `/host` route to fail. 

Draft PR where this new API would be used: https://github.com/viamrobotics/viam-apps/pull/45/files#r2116509944


jr22:
> just for my own understanding, we need this API to be able to retrieve a machine's ID in the case that the user/client doesn't directly provide it so that we can get the proper credentials?

yes exactly. so we know which API key to get. a bit indirect (since we could add an API for from name + locationid, get API key) but this feels more reusable

PR #703, opened 2025-06-03
[APP-8317] Add FragmentID Filter to ListMachineSummaries

JosephBorodach:
**See ticket [here](https://viam.atlassian.net/browse/APP-8317)**


michaellee1019 (changes requested):
Lets do a `repeated string fragment_ids` because there will be use cases where we'd want results for more that one fragment at a time, and doing that client side is painful.

I also looked back at the meeting notes (I wasn't there) but the requirements were documented as "Eventually this should be a cloud feature that shows me all machines that I have access to with a specific fragment. Pass list of ids."

michaellee1019 (approved):

PR #708, opened 2025-06-13
RSDK-10642 - add tcp-mode to module configs

stuqdog:


stuqdog:
cc @bhaney @penguinland

PR #720, opened 2025-07-09
Change geometry representation in proposed WorldStateStore scope

raybjork:


proto/viam/common/v1/common.proto
```diff
@@ -89,8 +89,8 @@ message Mesh {
   bytes mesh = 2;
 }
 
-message Points {
-  repeated float array = 1;
+message PointCloud {
```
raybjork, line 92: I could go either way on this name but this does seem consistent with our nomenclature

proto/viam/common/v1/common.proto
```diff
@@ -148,21 +148,21 @@ message GeoGeometry {
 // Transform contains a pose and two reference frames. `parent` is the starting reference frame, and `name` is the observer reference frame.
 // `pose` represents the pose of an object in the `parent` frame as observed within the `name` frame.
 message Transform {
-  // The name of the parent reference frame
-  string parent = 1;
+  // The name of the transform
+  string name = 1;
```
raybjork, line 152: this was backwards in the initial PR.  Only caught it on a closer read

biotinker (approved):
LGTM, consider adding the TODOs

PR #723, opened 2025-07-17
DATA-4345 Add new billing SKU to protos

vijayvuyyuru:


proto/viam/app/v1/billing.proto
```diff
@@ -71,6 +71,7 @@ enum UsageCostType {
   USAGE_COST_TYPE_TRAINING_LOGS_EGRESS = 21;
   USAGE_COST_TYPE_TABULAR_DATA_DATABASE_CLOUD_STORAGE = 22;
   USAGE_COST_TYPE_TABULAR_DATA_DATABASE_COMPUTE = 23;
+  USAGE_COST_TYPE_BINARY_DATA_CROSS_REGION_EGRESS = 24;
```
NickPPC, line 74: "CROSS_REGION"
Is it about the data moving between GCP regions within our GCP account? Or is it about data exiting the current GCP region it is in?

If a customer download all their data from GCP US east-2 to AWS us-east-1, is that this SKU ?

PR #727, opened 2025-07-23
RSDK-11268: Handle the prefix field for remotes

jmatth:


proto/viam/app/v1/robot.proto
```diff
@@ -293,6 +293,7 @@ message RemoteConfig {
 
   // Secret is a helper for a robot location secret.
   string secret = 10;
+  string prefix = 11;
```
benjirewis, line 296: [nit] Could you just document what this field does with a comment?

benjirewis (changes requested):

benjirewis (approved):

PR #728, opened 2025-07-28
Move tool management to mise

jmatth:
This moves management of the tools used to compile the protos to [mise-en-place][mise]. This accomplishes two things: 

- Pins the versions of all our build tools so we can reason about upgrading them
- Keeps local and CI installed tools in sync so that local and CI builds produce the same output

Some details about the move to mise:

- Wherever possible we have mise install prebuilt binaries
- For several tools there are no plugins in mise's registry for prebuilt binaries, so we use the experimental go backend to install directly from source
- Our lone js dependency has also been moved to be installed via mise's npm backend
- The gh actions to install + setup go and node have both been replaced w/ the mise action
- The mise gh action caches the tools it installs. The cache is quite large (~250MB) but cuts tool installation from 2 minutes to 4 seconds on a hit
- All the make targets have been moved to mise tasks. Doing this means developers can execute those tasks using just the mise binary without having to modify their shell environment

Example PR demonstrating the compile protos workflow [here][pr].

[mise]: https://mise.jdx.dev/
[pr]: https://github.com/jmatth/api/pull/4

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -29,7 +29,9 @@ node = "22"
 "npm:ts-protoc-gen" = "0.15.0"
 protobuf = "3.20.2"
 protoc-gen-go = "1.35.1"
-protoc-gen-grpc-web = "1.5.0"
+# The default asdf backend for this downloads the wrong binary on aarch64 linux.
+# Use ubi to grab the correct binary directly from the Github release.
+"ubi:grpc/grpc-web" = { version = "1.5.0", rename_exe = "protoc-gen-grpc-web" }
 
 [tasks.clean]
 description = "Remove all untracked files"


abe-winter:
is this a draft or ready for review?

.github/workflows/compile_protos.yml
```diff
@@ -44,17 +44,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           # note: check out fork because we will need it when we push below
           repository: ${{ github.event.pull_request.head.repo.full_name }}
-      - uses: actions/setup-node@v4
-      - uses: actions/setup-go@v5
```
abe-winter, line 48: getting rid of the setup-node + setup-go actions has caching implications

if builds are substantially slower, would try adding these back

.github/workflows/compile_protos.yml
```diff
@@ -64,16 +54,19 @@ jobs:
         if: startsWith(env.commit-message, env.CI_COMMIT_MESSAGE_PREFIX) && env.commit-author == env.CI_COMMIT_AUTHOR
         run: echo "is-auto-commit=true" >> $GITHUB_ENV
 
+      - uses: jdx/mise-action@v2
```
abe-winter, line 57: can you pin mise-action to a sha? unpinned 3rd party actions are a security issue

.github/workflows/compile_protos.yml
```diff
@@ -64,16 +54,19 @@ jobs:
...
         if: ${{ !env.is-auto-commit }}
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-      - uses: arduino/setup-protoc@v1
-        if: ${{ !env.is-auto-commit }}
 
       - name: Compile protos
         if: ${{ !env.is-auto-commit }}
-        run: make dist/buf
+        run: 'mise r --force clean ::: buf'
```
abe-winter, line 69: expand `r` to `run` imo -- makes it more accessible for whoever touches this next

abe-winter (approved):
lgtm if otterverse's questions are solved, but can you address caching + sha comments

.github/workflows/compile_protos.yml
jmatth: Sure. I'll move the buf action to use a sha while I'm at it unless there's a reason not to. The only reason I went with the version instead of a sha here is because that's how the other 3rd party actions (buf and arduino/setup-protoc) were being used.

Makefile
```diff
@@ -1,53 +1,32 @@
...
-export PATH := $(GOBIN):$(shell npm root)/.bin:$(PATH)
-
 .PHONY: all
-all: clean dist/buf
+all: print_deprecation
+	mise r clean ::: buf
+
+.PHONY: print_deprecation
+print_deprecation:
+	@echo "Makefile deprecated; please call mise run directly"
```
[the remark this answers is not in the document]
jmatth: One other thought: we could revert the changes to the makefile but leave the mise targets in place. Then users who are used to running make won't need to install mise yet but anyone in the know can test mise as a replacement task runner. It would denormalize things a bit but the intent isn't to keep both tools around forever.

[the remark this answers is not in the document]
dgottlieb: I'm with @Otterverse here. I think moving to mise to manage the backend is the right direction.

It may be the case where we want mise to be the frontend too. But I don't think this PR needs to simultaneously:
* become the backend and
* deprecate the frontend

Especially since the deprecations aren't saying what one ought to run. And some of the nouns are different (for the better, but different). E.g: dist/tool-install -> mise install (not `mise run` as the deprecation mentions) or dist/buf-go -> mise run buf-go (not dist/buf-go).

dgottlieb (approved):

jmatth:
@Otterverse I confirmed that the current branch works for me from a clean state (mise not installed, `~/.local/share/mise` and `~/.config/mise` removed, `go` and `node` not installed through brew)

etc/setup.sh
```diff
@@ -0,0 +1,19 @@
...
+if type brew > /dev/null; then
+  brew install mise
+elif type pacman > /dev/null; then
+  sudo pacman -Sy mise
+elif type apt > /dev/null; then
+  sudo bash -c 'apt update && apt install -y mise'
+elif type dnf > /dev/null; then
+  sudo bash -c 'dnf copr enable jdxcode/mise && dnf install mise'
+else
+  echo "Don't know how to manage packages on this system. Please install mise-en-place manually."
```
Otterverse, line 17: Should do the direct install here (the curl pipe to sh thing.) Also catch cases where mise is not found in the package managers.

Otterverse (changes requested):
So this works nicely on MacOS now, especially if I've already got tools like make and git setup/installed.

But... it still doesn't work in a clean/new/linux environment. I'm using a plain "debian" docker container as a test target in the below blobs.

```
testbot@33064ffa49a0:/host/api$ make
./etc/setup.sh
./etc/setup.sh: line 3: type: mise: not found
./etc/setup.sh: line 8: type: brew: not found
./etc/setup.sh: line 10: type: pacman: not found
./etc/setup.sh: line 13: sudo: command not found
make: *** [Makefile:13: dist/tool-install] Error 127
```
Took me a minute to realize most of those errors are "expected" and only the lack of sudo was the real problem. So installed (and configured sudo) as well. Suppressing expected errors in the "type" calls would be good.

```
testbot@33064ffa49a0:/host/api$ make
./etc/setup.sh
./etc/setup.sh: line 3: type: mise: not found
./etc/setup.sh: line 8: type: brew: not found
./etc/setup.sh: line 10: type: pacman: not found
Hit:1 http://deb.debian.org/debian bookworm InRelease
Hit:2 http://deb.debian.org/debian bookworm-updates InRelease
Hit:3 http://deb.debian.org/debian-security bookworm-security InRelease
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
All packages are up to date.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package mise
make: *** [Makefile:13: dist/tool-install] Error 100
```

Now it seems "mise" isn't available in the default repo for debian, so that's a non-starter too. Is there a reason not to just use the normal install method from the mise docs? That is one of the install+activate scripts from https://mise.jdx.dev/installing-mise.html#shell-specific-installation-activation ?

Ex: `curl https://mise.run/bash | sh`

I've found, of course, that I also need to manually install `make`, `gpg`, `curl`, and `git` for most stuff in mise to actually work too. Only once I did all that was it finally able to run. I think it's fine if these are "expected" to be installed by the user already or something, but they at least all need to be documented.

jmatth:
> curl https://mise.run/bash | sh

I have a general aversion to the "pipe this url into bash" install method but if we think it's secure enough I'm fine with it.

>  I think it's fine if these are "expected" to be installed by the user already or something, but they at least all need to be documented.

I think that's the most reasonable thing to do. Though at that point, is there a reason to have this whole automagic mise setup as part of the repo? Why not just say "we use mise, the install instructions are over here, come back when you're done and it should Just Work™️"? In the current setup we already assume at least make to run the `dist/tool-install` target, git to clone the repo in the first place, I'm pretty sure rdk assumes brew is available on MacOS, etc. We could just add mise to that list of tools users are expected to have available in order to bootstrap the dev environment.

PR #731, opened 2025-08-05
bump x/oauth2

abe-winter:
(For https://github.com/viamrobotics/api/security/dependabot/32).


jmatth (approved):
Might be worth jumping straight to [the latest version][1] as long as you're here but eh 🤷. LGTM 

[1]: https://github.com/golang/oauth2/releases

PR #740, opened 2025-08-12
Work around weird buf path handling, update buf configs

jmatth:
I'm bundling a couple changes together here that aren't really related to avoid having to open multiple PRs:

- Fix buf using `protoc-gen-js` from path instead of the `protoc`'s builtin js compiler. This would cause differing js proto output on machines where users had `protoc-gen-js` available in their path.
- Update the buf config files to v2 and update a string in `buf.yaml` that was causing buf to print a deprecation warning. I initially thought the v2 config format might help solve the previous point but even though it didn't we might as well keep the updated config file format.


Otterverse (approved):
LGTM, just ask that you keep a close eye on actions here for a bit (which you're probably already doing) in case anything starts causing issues for others.

PR #745, opened 2025-08-15
Add DNS names to Machine and Part summaries

NickPPC:
https://viam.atlassian.net/browse/APP-9174

### DIFF
--- a/proto/viam/app/v1/app.proto
+++ b/proto/viam/app/v1/app.proto
@@ -1160,6 +1160,7 @@ message LocationSummary {
 message MachineSummary {
   string machine_id = 1;
   string machine_name = 2;
+  optional string main_part_dns_name = 4;
   repeated PartSummary part_summaries = 3;
 }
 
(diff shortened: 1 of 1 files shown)


proto/viam/app/v1/app.proto
```diff
@@ -1160,6 +1160,7 @@ message LocationSummary {
 message MachineSummary {
   string machine_id = 1;
   string machine_name = 2;
+  optional string main_part_dns_name = 4;
```
cheukt, line 1163: i think it makes sense to surface here, but then we should also add main_part_name here as well

NickPPC: Adding the main part *name* feels wrong. We could add the main part ID if needed
Or we can just add a `is_main_part` boolean on the PartSummary

NickPPC: @dgottlieb `main_part_dns_name` was intended to be the string you use to dial. We have a `dns_name` on the `Part` proto struct.
I am open to alternative names:
- `main_part_hostname`
- `main_part_url`
- `main_part_dial_name`

cheukt: with `is_main_part` I don't think we need to add `main_part_dns_name`. tho not sure how users are most likely to want this data

PR #749, opened 2025-08-25
RSDK-11264 - deprecated resourceName fields and added corresponding string fields and 

dhritinaidu:
For all motion API defined in [motion.proto](https://github.com/viamrobotics/api/blob/0dc6388c26fb0d0a40f84d82a78b6532525c9ac6/proto/viam/service/motion/v1/motion.proto), deprecates any field foo of type ResourceName and calls that field foo_deprecated, and adds a corresponding, new string field named foo instead.


benjirewis (approved):
This seems correct to me; thanks! Let's hold off on merging until https://github.com/viamrobotics/rdk/pull/5221 is complete. A note that once you merge this, upgrading `api` in any of the SDKs or `rdk` will cause issues, so we'll have to do those corollary changes _somewhat_ soon after this merge.

benjirewis:
Ah gotcha, yeah if you guys can verify that maintaining the same numerical value and only changing the name of the field continues to allow correct [de]serialization, that works for me. Thanks for bringing it up, @stuqdog .

PR #838, opened 2026-04-14
[APP-16031] Add unlisted favorite DoCommand APIs

martha-johnston:
https://viam.atlassian.net/browse/APP-16031

This PR adds the necessary APIs to add, edit, remove, and list user favorite DoCommands. There won't be SDK wrappers APIs so it will be for internal use.

Scope doc [here](https://docs.google.com/document/d/1ytkcsrZex1qlBCXIi-oH8cUI3Tcjb0gnyaypE2mz6nE/edit?tab=t.0#heading=h.tcicyojyqi6c)


martha-johnston:
@gmulz @DTCurrie @clintpurser I took this over from Kevin so please let me know if I missed anything but I think this covers it

gmulz (changes requested):
these are not "unlisted", everything in `api` is a public api and this is a public repo

PR #840, opened 2026-04-26
APP-14947: add last_payment_failure_{reason,at} to InvoiceSummary

n0nick:
Surface the payment failure category recorded on outstanding invoices by viamrobotics/app#11777 (merged) so clients can render category-specific copy and CTAs (update card, retry, contact issuer, etc.).

Two additive fields on `InvoiceSummary`:
- `last_payment_failure_reason` — documented closed-set string
- `last_payment_failure_at` — `google.protobuf.Timestamp`


proto/viam/app/v1/billing.proto
```diff
@@ -38,6 +38,15 @@ message InvoiceSummary {
   string status = 4;
   google.protobuf.Timestamp due_date = 5;
   google.protobuf.Timestamp paid_date = 6;
+
+  // Category of the most recent payment failure. One of: card_declined,
+  // insufficient_funds, expired_card, incorrect_cvc, incorrect_card_number,
+  // authentication_required, fraud_suspected, processing_error, other.
+  // Empty when no failure is recorded. Clients must tolerate unknown values.
+  string last_payment_failure_reason = 7;
```
edobranov, line 46: [q] Will this arbitrarily prioritize one of `LastStripeFailureCode` or `LastStripeDeclineCode` from the `Invoice` struct? Totally fine if so, perhaps worth a quick note here if you think it's relevant to mention (also fine to rely on the implementation to show this)

edobranov (approved):

proto/viam/app/v1/billing.proto
n0nick: not arbitrary, `LastStripeDeclineCode` is used when `LastStripeFailureCode` is `card_declined` (DeclineCode essentially a sub-type of FailureCode). I wasn't sure if I should document this here, as the proto and comments are public and Stripe is an implementation detail.

edobranov: Got it, no need to mention then

PR #848, opened 2026-05-08
Add builder flag and fallback message to StartBuild proto (APP-16314)

cecilyViam:
## Summary

- Adds optional `builder` field (field 9) to `StartBuildRequest` — callers pass `viam-cloudbuild-test` to target the new Viam Linux build service
- Adds optional `builder_fallback_message` field (field 2) to `StartBuildResponse` — server uses this to communicate when an org is not whitelisted and the default builder was used instead

## Context

Part of APP-16314 (Builder Target Flag). This proto change is a prerequisite for the app and rdk PRs — both need to update their `go.viam.com/api` dependency to pick up these fields.

## Test plan

- [ ] Proto regenerated via `mise run buf-go` — rawDesc updated, struct fields and getters generated correctly
- [ ] app and rdk PRs compile and test against this version


amaschas (approved):
🐴

PR #850, opened 2026-05-13
RSDK-13966 Add PlayStream to audio out 

oliviamiller:
add playstream RPC to audio out 

Scope: https://docs.google.com/document/d/1j3HqaDqhk1mVqKixENFzowqhgbtsjFvXtl8IejE3pHw/edit?tab=t.0


proto/viam/component/audioout/v1/audioout.proto
```diff
@@ -15,6 +15,8 @@ service AudioOutService {
   rpc Play(PlayRequest) returns (PlayResponse) {
     option (google.api.http) = {post: "/viam/api/v1/component/audioout/{name}/play"};
   }
+  // PlayStream streams audio chunks to the audioout component for playback.
```
seanavery, line 18: Might be worth expanding the rpc comment here a bit to include init and error behavior

proto/viam/component/audioout/v1/audioout.proto
```diff
@@ -44,4 +46,23 @@ message PlayRequest {
...
 
+message PlayStreamRequest {
+  oneof payload {
+    PlayStreamInit init = 1; // sent on first message
+    PlayStreamChunk audio_chunk = 2;
+  }
+}
+
+message PlayStreamChunk {
+  bytes audio_data = 1;
```
seanavery, line 63: [question] I see this matches up with the scope doc, but, is there a reason we are just going with raw bytes here and not including timestamp and sequence metadata that is found in AudioChunk?

seanavery (approved):
LGTM. Matches up with scope doc well


`google.api.http` is in the scope doc but not sure if it makes sense with the streaming endpoint here. Could be worth updating the scope

proto/viam/component/audioout/v1/audioout.proto
oliviamiller: The metadata isn't needed for any current use cases with this API, we can add metadata if needed for the future. See this [comment thread](https://docs.google.com/document/d/1j3HqaDqhk1mVqKixENFzowqhgbtsjFvXtl8IejE3pHw/edit?disco=AAAB6You-pA) on scope for more context.

oliviamiller:
> LGTM. Matches up with scope doc well
> 
> `google.api.http` is in the scope doc but not sure if it makes sense with the streaming endpoint here. Could be worth updating the scope

thanks for catching i meant to add the http, other streaming RPCs include it so will follow that pattern.

PR #851, opened 2026-05-18
APP-16462: Add DatasetType to dataset proto

vpandiarajan20:
## Summary

Adds a `DatasetType` enum (`BINARY_DATA`, `SEQUENCE_DATA`) to the dataset proto, plus:
- `Dataset.type` field on the `Dataset` message (number 5)
- `optional DatasetType type` field on `CreateDatasetRequest` (number 3)

`optional` on `CreateDatasetRequest.type` lets the server distinguish "client didn't send a type" from "client explicitly sent the zero value" — both currently map to `BINARY_DATA` for backwards compatibility, but the explicit-unset signal is useful for telemetry and future behavior.

Prerequisite for the sequence training datasets work tracked in epic [APP-16329](https://viam.atlassian.net/browse/APP-16329) — App-side handler changes follow in a separate PR once this lands and Go stubs are regenerated.

## Test plan

- [x] Regenerate Go + JS stubs with the pinned toolchain
- [x] Verify generated `Dataset.GetType()` and `CreateDatasetRequest.GetType()` accessors exist with the expected types
- [x] App-side follow-up PR consuming the new field

[APP-16329]: https://viam.atlassian.net/browse/APP-16329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vpandiarajan20:
Noting that we are using `DATASET_TYPE_BINARY_DATA = 0` + inline `buf:lint:ignore`

njooma (approved):

PR #857, opened 2026-05-19
Revert "RSDK-9718: fix tagger.proto go_package so its descriptor registers correctly"

ale7714:
Reverts viamrobotics/api#847


ale7714:
@NickPPC i'm making a patch release for rdk because this is breaking the rstp modules.

PR #859, opened 2026-06-01
APP-14575 Support Deprecation

enternal-L:
(https://viam.atlassian.net/browse/APP-16741)

## PR Stack
- **You are here ->** https://github.com/viamrobotics/api/pull/859
- Go changes: https://github.com/viamrobotics/app/pull/12273 
- Registry page: 
     - Individual: https://github.com/viamrobotics/app/pull/12274
     - Search: https://github.com/viamrobotics/app/pull/12295
- Machine page: https://github.com/viamrobotics/app/pull/12303
- Fragment page: https://github.com/viamrobotics/app/pull/12313

## Goal
Add module and version deprecation feature to protos

## Changes 
- Added `DeprecatedStatus` that stores optional message and data of deprecation
- Added new Read Protos(contains `DeprecatedStatus`)
  - `ModuleVersion`
  - `RegistryItem`
  - `VersionHistory`
- Added new Write Protos
  - Deprecate/Undeprecated requests & responses

## Testing Guidance
- Testing is done on subsequent PRs that deal with the frontend interaction ex. Undeprecating and Deprecating a module and module version should work and reflect on the page correctly

## Notes
- The existence of `DeprecatedStatus` means that item is deprecated, the absence of it means the item is undeprecated.
- Both `ModuleVersion` and `VersionHistory` must hold `DeprecatedStatus` since `ModuleVersion` is the proto that  interfaces from the DB, and `ModuleVersion` later gets converted into`VersionHistory`, surfacing to the frontend
- All requests and responses are its own struct, rather than a field in existing structs that writes `ModuleVersion`, `RegistryItem`, and `VersionHistory`

## Tasks
- Add deprecation of registry item, module, and version to protos ✅ 
- Add undeprecation of registry item, module, and version to protos ✅


michaellee1019:
@enternal-L @petrnovikov I think this implements more APIs that we discussed. I thought we were going to implement just `DeprecateRegistryItem`, `UndeprecateRegistryItem`,  `DeprecateRegistryItemVersion`, and `UndeprecateRegistryItemVersion`?

I think we can omit the Module named APIs. This way in the future we could implement deprecation for ML models without any API changes. Historically we have Module named APIs because they existed before ML models were in the registry. Duplicate APIs that had a type field were added and named as RegistryItem. So going forward I'd like to not implement any new Module level APIs.

proto/viam/app/v1/app.proto
```diff
@@ -1814,6 +1857,34 @@ message ListModulesResponse {
...
+message UndeprecateModuleRequest {
+  string module_id = 1;
+}
+
+message UndeprecateModuleResponse {}
+
+message DeprecateModuleVersionRequest {
+  string module_id = 1;
+  string version = 2;
+  optional string message = 4;
```
edobranov: Let's number these wire tags 1, 2, 3. It's functionally fine to skip a number, but it can confuse someone in the future thinking there was a 3 that was removed. In that case, they're not supposed reuse 3 since it could cause a misinterpretation of bytes from an old client.

edobranov (dismissed):

petrnovikov:
> @enternal-L @petrnovikov I think this implements more APIs that we discussed. I thought we were going to implement just `DeprecateRegistryItem`, `UndeprecateRegistryItem`, `DeprecateRegistryItemVersion`, and `UndeprecateRegistryItemVersion`?
> 
> I think we can omit the Module named APIs. This way in the future we could implement deprecation for ML models without any API changes. Historically we have Module named APIs because they existed before ML models were in the registry. Duplicate APIs that had a type field were added and named as RegistryItem. So going forward I'd like to not implement any new Module level APIs.

Seconding Michael's point: `DeprecateModule`, `UndeprecateModule`, `DeprecateModuleVersion` and `UndeprecateModuleVersion` are unnecessary RPCs.

proto/viam/app/v1/app.proto
```diff
@@ -1597,6 +1623,19 @@ message TransferRegistryItemRequest {
 
 message TransferRegistryItemResponse {}
 
+message DeprecateRegistryItemRequest {
+  string item_id = 1;
+  optional string message = 2;
+}
+
+message DeprecateRegistryItemResponse {}
```
petrnovikov, line 1627: Should we add some sort of response? Like `success` or `status` boolean to be able to ensure that deprecation went successfully on the frontend?

proto/viam/app/v1/app.proto
```diff
@@ -1597,6 +1623,19 @@ message TransferRegistryItemRequest {
...
+
+message UndeprecateRegistryItemRequest {
+  string item_id = 1;
+}
+
+message UndeprecateRegistryItemResponse {}
```
petrnovikov, line 1633: Also, here we should probably add some status indicator in the response.

proto/viam/app/v1/app.proto
enternal-L: resolved

proto/viam/app/v1/app.proto
petrnovikov: Disregard, existing pattern proves me wrong.

proto/viam/app/v1/app.proto
petrnovikov: Disregard, existing pattern proves me wrong.

michaellee1019 (approved):

PR #861, opened 2026-06-02
Add filterdatasetsbytype

njooma:
This will be useful to filter based off of type from the backend, rather than requesting all datasets for an org and filtering based on type in the frontend. 

Other consideration -- updating `ListDatasetsByOrganizationIDRequest` to include a `type`. ~Originally decided against it because it requires changing existing server-side code, but I don't feel strongly either way and can happily switch to that.~ **This ended up being the preferred implementation**


vpandiarajan20:
I'm in favor of modifying ListDatasetsByOrganizationIDRequest, seems like a low lift although it is a breaking change.

vpandiarajan20 (approved):

etai-shuchatowitz (approved):

PR #865, opened 2026-06-11
[APP-16897] Modify API to support typed fragment variables (map<string, string> → Struct)

jckras:
(https://viam.atlassian.net/browse/APP-16897) - This PR modifies the `FragmentImport` message in the app API, adding a `fragment_variables` field and deprecating the existing `variables` field.

This needed to change because `map<string, string>` forces every variable value to be a string. Org default fragment variables travel through this field, and the should be able to carry any type. We are using `Struct` now bc it carries any JSON value, and it's what the fleetdashboard `FragmentImport` already uses for these same variables.

By adding a new field and deprecating the old, this is a NON-breaking change :) 

**Follow ups:**
app switches to `fragment_variables`, still populating the old field for older clients (app PR ready: viamrobotics/app#12333)


jckras:
@gmulz @tgritter still trying to figure out how to determine what the blast radius of this change is, aka figure out "how" breaking it is

proto/viam/app/v1/app.proto
```diff
@@ -1047,7 +1047,9 @@ message FragmentImport {
   string fragment_id = 1;
   string version = 2; // revision or tag
   optional string prefix = 3; // namespace prefix
-  map<string, string> variables = 4; // key-value pairs for fragment variables
+  // Deprecated: use typed_variables
+  map<string, string> variables = 4 [deprecated = true]; // key-value pairs for fragment variables
+  optional google.protobuf.Struct typed_variables = 5;
```
gmulz: since the variables should have been typed to begin with, this sounds a little awkward. maybe `fragment_variables` is good?

jckras: good idea, better name

petrnovikov (approved):
I'm not very familiar with the context, but since this change doesn't delete any fields and adds an optional field, it can't break anything. Hence, approving!

michaellee1019 (approved):

gmulz (approved):

PR #869, opened 2026-06-17
APP-16744: Add boolean for filter search deprecation

enternal-L:
## Goal
Upon [discussion](https://github.com/viamrobotics/app/pull/12295#discussion_r3417118495), deprecation filtering on registry search will be delegated to the backend rather than the frontend i.e. the toggle button sends a query request rather than filtering on items in the frontend


proto/viam/app/v1/app.proto
```diff
@@ -1592,6 +1592,8 @@ message ListRegistryItemsRequest {
   optional bool include_markdown_documentation = 9;
   repeated ModuleSourceType module_source_types = 10;
   repeated ModuleLanguage module_languages = 11;
+  // for "show deprecated" toggle button, showing owned deprecated items
+  bool show_owned_deprecated = 12;
```
petrnovikov: Should this be optional defaulting false?

petrnovikov (approved):
LGTM since the field is optional and should not cause breaking changes, that said, please get a second approval before merging.

proto/viam/app/v1/app.proto
```diff
@@ -1592,6 +1592,8 @@ message ListRegistryItemsRequest {
   optional bool include_markdown_documentation = 9;
   repeated ModuleSourceType module_source_types = 10;
   repeated ModuleLanguage module_languages = 11;
+  // for "show deprecated" toggle button, showing owned deprecated items
```
petrnovikov: Nit: but my preference would be to remove the comment (I know there are two other comments in this proto), naming is self explanatory.

michaellee1019 (approved):
I tried to think of a better name but this is fine. In the future we could add a `show_all_deprecated` or `show_unowned_deprecated` and it would fit next to this fine.

PR #876, opened 2026-07-07
APP-16998: Add Multi-select delete platforms for versions

enternal-L:
(https://viam.atlassian.net/browse/APP-16998)

## Goal
For module versions, when deleting platforms, I would like to support multi-select on the modal rather than having to delete one by one.


proto/viam/app/packages/v1/packages.proto
```diff
@@ -76,7 +76,9 @@ message CreatePackageResponse {
 message DeletePackageRequest {
   string id = 1;
   string version = 2;
-  PackageType type = 3;
+  PackageType type = 3 [deprecated = true];
```
michaellee1019: I think don't put the deprecated flag here. Its interpreted differently in each language but ends up failing builds in some languages.

We'll probably need to keep it around in the long term, so its fine to have both and still support it. For example older versions of the python SDK will reference it https://python.viam.dev/autoapi/viam/gen/app/packages/v1/packages_pb2/index.html#viam.gen.app.packages.v1.packages_pb2.DeletePackageRequest. And we currently don't have a great way of deprecating SDK versions. Some time in the future we might have a real deprecation process.

michaellee1019: Instead I would put a comment:
// Deprecated: use types instead.

michaellee1019 (approved):

PR #877, opened 2026-07-09
APP-16894: Add min_viam_server_version to registry item protos

jckras:
(https://viam.atlassian.net/browse/APP-16894) - This PR adds an optional `min_viam_server_version` string (semver, ex "0.62.0") to the registry-item API surface.

- `ModuleMetadata`: the module’s current value. This is what GetRegistryItem returns for a module. The registry page displays it from here, and the machine-config search banner (which I will be adding) will compare against it when you’re adding a module (new adds default to latest, so the current value is the right one to check).

- `ModuleVersion`: the frozen per-version value. ModuleMetadata contains a list of these. One per published version. Since each version freezes the value at publish time, the module card banner (which I will be adding) will read the specific configured version’s value from here. 

- `UpdateModuleMetadata`: how the value gets set. When a module owner edits their module on the registry page, the frontend sends an UpdateRegistryItemRequest, and module-specific edits travel inside this message. Adding the field here gives the registry page a way to write it. 

Additive and non-breaking. No changes to module APIs (UpdateModuleRequest, VersionHistory, etc.) the CLI is untouched.

This app PR ([APP-16895](https://viam.atlassian.net/browse/APP-16895)) is built against this branch so we can more easily test these api changes.

[APP-16895]: https://viam.atlassian.net/browse/APP-16895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


michaellee1019 (approved):

edobranov (approved):

jckras:
succesfully tested these api changes. see the pr description in this APP pr for proof: https://github.com/viamrobotics/app/pull/12704

PR #882, opened 2026-07-22
APP-17538: return per-file URLs from sequence dataset export

vpandiarajan20:
Adds `repeated SequenceDatasetExportFile files` to `GetSequenceDatasetExportResponse` and removes the single `download_url`. 
The export produces three independent parquet files, and zipping them forces serialization into one stream, so app currently buffers two of them fully in memory while the third is written: [`domains/dataset/export.go#L61-L88`](https://github.com/viamrobotics/app/blob/844583faf1eb83f2ecb6726a812ff625835f4d62/domains/dataset/export.go#L61-L88) (internal repo) shows the three generators each filling a `bytes.Buffer` before the zip is assembled. 


vpandiarajan20:
Switched from removing `download_url` to deprecating it.

Removing it forces a hard cutover — app and the CLI can't deploy atomically, so any window where a new server meets an older `viam` CLI breaks every export (the CLI downloads `GetDownloadUrl()` to `<datasetID>.zip`). Keeping the field deprecated lets the server emit both the zip and the per-file URLs during the transition, then drop it once clients have migrated.

- `string download_url = 3 [deprecated = true];` restored, `reserved 3` / `reserved "download_url"` dropped
- `files = 8` unchanged
- Regenerated Go and web bindings; diff is limited to the four dataset files
- `buf lint` clean; `buf breaking` reports nothing for `dataset.proto` (the four it flags are pre-existing drift in `app.proto` and `data_pipelines.proto` on main)

Generation note for anyone rerunning this: `mise r buf-web` must run with PATH containing **only** mise binaries, as the task comment says. With the system PATH appended, a different js generator wins and rewrites all ~40 `gen/js` files (`globalThis`, indentation). Worth knowing if you see an unexpectedly large diff.

vpandiarajan20:
Closing — the OOM fix took a different shape and this is no longer needed.

viamrobotics/app#13155 keeps the zip and streams each parquet directly into its zip entry, sequentially. That removes the `bytes.Buffer`s (the actual cause of the OOM) without changing the response contract, so `download_url` stays a single URL to a single zip and no `files` field is required.

Sequential generation turned out to cost almost nothing — `tabular_data.parquet` dominates the output, so serializing the three generators measured 31.3s vs 32.5s for the concurrent three-object version. Verified end to end: the dataset that OOM-kills unfixed staging at 40s now exports 2.74 GiB in 237s at 11.2% of a 4 GiB limit.

The proto work here is good and reviewed-clean, so it's worth keeping in mind rather than discarding. Reopen if we want direct-readable parquet in GCS — a training job or query engine can read a parquet in place with column pruning and predicate pushdown, but not one inside a zip, which has to be downloaded and unpacked first. That would justify the three-object contract on its own merits, independent of memory.

Tracked in APP-17538 (closed as obsolete, with this rationale) should we revisit.

PR #883, opened 2026-07-23
[APP-16771] Add count to SequencesByDatasetIDResponse

njooma:
So that we can appropriately display the number of total sequences in a dataset, even when paginated


vpandiarajan20:
Would this return the full number of sequences in the dataset? 
- If so, then this would be another database query, right? 
- If not, then we'd have to call this endpoint multiple times to get the full number.

Why don't we add another endpoint, like we did for binary data datasets?

njooma:
@vpandiarajan20 the goal is to have it return total count.

I don't think we have a separate endpoint for binary datasets? `BinarayDataByDatasetID` is internal (can see it in App but not API). The request for that does allow users to request `countOnly`, though, maybe we do that too? Or I could just write a new endpoint internally only

vpandiarajan20:
> Vignesh P (@vpandiarajan20) the goal is to have it return total count.
> 
> I don't think we have a separate endpoint for binary datasets? `BinarayDataByDatasetID` is internal (can see it in App but not API). The request for that does allow users to request `countOnly`, though, maybe we do that too? Or I could just write a new endpoint internally only

You're totally right, I think countOnly makes sense. Let's do that. I think you need to send out a quick scope email because its a public API change?

njooma:
> I think you need to send out a quick scope email because its a public API change?

I usually don't if it's additive/non-breaking, but maybe I'm supposed to be...

vpandiarajan20 (approved):

PR #887, opened 2026-07-30
[APP-16884] Add GetSubsystemVersionStatus to AgentDeviceService

danielbotros:
## Description

[APP-16884](https://viam.atlassian.net/browse/APP-16884) Quickly deprecate/pull a viam-server/agent version

Adds `GetSubsystemVersionStatus(subsystem, version) -> {deprecated}` to `viam.app.agent.v1.AgentDeviceService`. Devices and the machine config UI use it to warn when a machine is running or pinned to a deprecated version.

[APP-16884]: https://viam.atlassian.net/browse/APP-16884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cheukt:
will the machine ui call this endpoint repeatedly? conceptually i like the idea but putting it on the agent service feels kinda wrong. if we're just limiting the calls to the UI/app backend, can we just use the existing internal apis instead?

PR #888, opened 2026-07-31
[APP-17427] Add `order` and `range` fields to `GetRobotPartLogsRequest`

edobranov:
https://viam.atlassian.net/browse/APP-17427

## Where this sits in the sequence

Part of the logs CLI modernization. Six changes total, bringing the CLI to parity with the redesigned logs page now that reads are moving to ClickHouse. **This PR is 3.**

| # | Repo | Change | Ticket | PR |
|---|---|---|---|---|
| 1a | app | Page size 100 → 1000; default window 12h → 24h | APP-17245 | https://github.com/viamrobotics/app/pull/13004 |
| 1b | rdk | Remove CLI log count caps; `--count` becomes an opt-in cap | APP-17245 | https://github.com/viamrobotics/app/pull/13004 |
| 2 | app | Logs-page export command always emits `--start`/`--end` | APP-17426 | https://github.com/viamrobotics/app/pull/13050 |
| **3** | **api** | **Add `order` + `range` to `GetRobotPartLogsRequest` protos** | **APP-17247** | **◀ this PR** |
| 4 | app | ClickHouse query support for `order` + `range` | APP-17428 | TBD |
| 5 | rdk | `--order` / `--range` flags on `machines logs` | APP-17429 | TBD |
| 6 | app | Export command emits `--order=ascending` + `--range` for relative filters | APP-17430 | TBD |

Ordering: 3 gates everything after it — both app and rdk consume `go.viam.com/api` as a versioned module, so the proto change must land *and* be released before 4 and 5 compile. Then 4 → 5 → 6 in sequence.

## Summary

Adds two optional fields to `GetRobotPartLogsRequest` so callers can control log sort order and select a time window relative to a single anchor instead of supplying both endpoints.

## Notes for reviewers

`range` is a `string` rather than `google.protobuf.Duration`. `Duration`'s text form is a seconds count (`"864000s"`), which loses the compact `10d`-style input this field is meant to accept. Two consequences for the server side: it must parse the string itself, and `10d` is not valid for Go's `time.ParseDuration`, so day handling needs to be explicit (or callers write `240h`).

The mutual-exclusion rule and the anchor resolution are documented in comments only — proto cannot express them — so validation belongs in the app-side handler. This PR is the wire contract; it does not implement the behavior.

`TailRobotPartLogsRequest` is intentionally unchanged.


proto/viam/app/v1/app.proto
```diff
@@ -908,6 +918,15 @@ message GetRobotPartLogsRequest {
   optional int64 limit = 8;
   optional string source = 9;
   optional bool user_facing_only = 10;
+  // logs are returned newest first when the order field is empty
+  optional LogOrder order = 11;
+  // range is a duration string (e.g. "10m", "10h", "10d") that is resolved
```
abe-winter: if this is parsed using golang duration parsing, can you say that in the docstring?

abe-winter (approved):
one comment nit but lgtm

proto/viam/app/v1/app.proto
edobranov: Technically it won't be a true Go duration parse because Go's parsing doesn't have the `d` unit. Also we'll want to disallow seconds down to nanoseconds. So I think it'll end up being a manual check for m,h,d then parsing it as a duration.

edobranov: but I'll clarify it can only be one of those three

abe-winter: fair enough

I always err on the side of using a standard, but you have thought about this more so your call

PR #889, opened 2026-08-20
[APP-17564] Add resource_subtype to SequenceResourceFilter

katiepeters:
Sequences identify their resources by `(resource_name, method_name)` only. Adding the subtype lets app return the full `(subtype, name, method)` triple on reads.

```proto
message SequenceResourceFilter {
  string resource_name = 2;
  string method_name = 3;
  // Populated on read. Optional on create/update: when provided it is validated
  // against captured data the same way an omitted value is resolved.
  string resource_subtype = 4;

  reserved 1;
  reserved "part_id";
}
```

Field `4` is the next free number — `1` is reserved from the removed `part_id`.

### Why it's optional on write

App resolves the subtype server-side from captured data, falling back to the machine config. Clients never need to send one, so no SDK or RDK change is required for this to work. When a client *does* send one, app validates it against captured data and rejects a mismatch rather than silently substituting — the likely mistake is the bare form (`camera`) where storage uses the triplet (`rdk:component:camera`).

### What it unblocks

- `viam data export sequence` (viamrobotics/rdk#6365) currently resolves the subtype per resource with a `limit: 1` `TabularDataByFilter`, noted there as "removable when sequence resources record subtypes". That cleanup also retires a deprecated call site.
- App-side consumers: viamrobotics/app `katie/APP-17564`.

### Note on the diff

Only `data.pb.go` is regenerated. `data_grpc.pb.go` was left untouched — adding a message field can't change service stubs, and regenerating it locally produced a 320-line diff purely from a `protoc-gen-go-grpc` version difference (1.5.1 local vs 1.2.0 pinned in `mise.toml`).


katiepeters:
Going a different route

PR #892, opened 2026-08-24
Register and delete custom containers - proto

lvhg:
https://viam.atlassian.net/browse/RSDK-14444


proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,6 +190,32 @@ message ListSupportedContainersResponse {
   // }
 }
 
+message RegisterCustomTrainingContainerRequest {
+  string organization_id = 1;
+
+  // name format will be [user inputted name]:[first 8 characters of org id] 
+  string name = 2;
```
katiepeters: Remind me why we need `:[first 8 characters of org id]`?

lvhg: It's to prevent possible collisions between custom containers and viam containers

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,6 +190,32 @@ message ListSupportedContainersResponse {
...
+  // name format will be [user inputted name]:[first 8 characters of org id] 
+  string name = 2; 
+
+  // Docker Hub reference, tag required: docker.io/<namespace>/<repo>:<tag>.
+  // Official images normalize to docker.io/library/<repo>:<tag>.
+  string image_uri = 3;
+
+  // Optional. Defaults to the image tag. Registered key is "name:version".
+  string version = 4; 
+  string description = 5; // will serve as display name
```
katiepeters, line 210: I know we've already touched on this in at least one meeting, but it still seems odd to be to use description as a name. Would we run into a scenario where a user added a fairly long description, not realizing it was really a name, or will we always be masking the name of the field from them (so that they'll think they're providing the display name)? Or how will that be handled

proto/viam/app/mltraining/v1/ml_training.proto
katiepeters: What kind of collision? How would such a collision come about

proto/viam/app/mltraining/v1/ml_training.proto
lvhg: I see what you mean, John didn't want us to change it though since description is an existing field and changing it could cause issues. I am thinking we can add help text in the CLI to clarify things

proto/viam/app/mltraining/v1/ml_training.proto
lvhg: The keys are name:version, so for example a collision would happen if a user had name=tf and version=2.16 since that matches one of the viam containers

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
+}
+
+message RegisterCustomTrainingContainerResponse {
+  Container container = 1;
+}
+
+message DeleteCustomTrainingContainerRequest {
+  string organization_id = 1;
+  string key = 2;
```
katiepeters: Not the id?

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
   //     created_on: 2026-01-15T01:30:15.01Z
   //     organization_id: "xyz"
   //     visibility: VISIBILITY_PUBLIC
+  //     created_on: 2026-01-15T01:30:15.01Z
+  //     organization_id: "xyz"
+  //     visibility: VISIBILITY_PUBLIC
```
katiepeters: Duplication?

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
   //   }
   // }
 }
 
+message RegisterCustomTrainingContainerRequest {
+  string organization_id = 1;
+
+  // On registration resolve URI to container's image:tag to set as key.
```
katiepeters: I don't think this comment is necessary?

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
+  // Docker Hub reference, tag required: docker.io/<namespace>/<image>:<tag>.
+  // Official images normalize to docker.io/library/<image>:<tag>.
```
katiepeters: I'm trying to make sense of these comments. Would it be true to say this:
```
// Docker Hub image reference. Must include a tag; references without one
// are rejected at validation (no implicit :latest).
//
// The registry prefix is optional on write and normalized on read:
//   <image>:<tag>              -> docker.io/library/<image>:<tag>
//   <namespace>/<image>:<tag>  -> docker.io/<namespace>/<image>:<tag>
```

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
+  string image_uri = 3;
```
katiepeters: Is there a reason `image_uri` is 3 and `description` is 5?

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
+
+  string description = 5; // will serve as display name
+}
+
+message RegisterCustomTrainingContainerResponse {
+  Container container = 1;
```
katiepeters: Returning the `id` might be the better convention?

proto/viam/app/mltraining/v1/ml_training.proto
```diff
@@ -183,10 +190,35 @@ message ListSupportedContainersResponse {
...
+}
+
+message DeleteCustomTrainingContainerRequest {
+  string organization_id = 1;
```
katiepeters: If deleting by id, we do not need to include `organization_id`

katiepeters (approved):

PR #902, opened 2026-09-22
Add new ListContainers RPC

lvhg:
This RPC is similar to ListSupportedContainers, but takes an org id parameter and will show containers from that org in addition to Viam containers. The response will be a list instead of a map.


lvhg:
duplicate

PR #901, opened 2026-09-22
[OLD/ignore] APP-18024: Add ListContainers RPC

katiepeters:
## Summary
- Adds a new `ListContainers` RPC to `MLTrainingService` that takes an `organization_id` and returns the available training containers (public containers + the org's custom containers) as a `repeated Container` list, instead of the `map<string, Container>` returned by `ListSupportedContainers`.
- `ListSupportedContainers` is left unchanged so older CLI versions keep working.

PR #903, opened 2026-09-22
New ListContainers RPC

lvhg:
This RPC is similar to ListContainers but includes an org ID parameter and will return containers from that org as well as Viam containers. It returns containers as a list.


katiepeters:
When we spoke this morning, you asked me to handle this, so I did: https://github.com/viamrobotics/api/pull/901

katiepeters (changes requested):
See comment

lvhg:
duplicate

